PR #1, opened 2022-01-31
KBV-142 Create Address /session API (slice 1)

idrop:
First slice of creating the Address CRI session API. 

This will be used by the Address CRI frontend to initialise a session with the API. Here's what the interaction looks like:
Screenshot 2022-01-31 at 12 31 17

- [x] Add basic gradle structure
- [x] Add API gateway with request validation
- [x] minimal lambda handler that just returns a UUID as a session id
- [x] Tested by deploying to `di-ipv-dev`


MattBidewell (approved):
Most of it is boiler code, which I cant say for certain is right.
But the session handler looks good

PR #2, opened 2022-02-08
KBV-142 Reference openapi definition from local file

idrop:
## Proposed changes

Reference the openapi yaml file from a local file, rather than s3

### What changed

Removed the need to place the openapi yaml in s3, we can refer to it locally.

### Why did it change

Using an S3 bucket makes more moving parts, and I didn't know how to do this previously - props to @jkunle for figuring this out.

### Issue tracking

- [KBV-142](https://govukverify.atlassian.net/browse/KBV-142)


MattBidewell (approved):
LGTM

PR #3, opened 2022-02-16
Tino store session

tinokingstone:
### What changed

The session id is now saved into a dynamo db table.

### Issue tracking

- [KBV-XXXX](https://govukverify.atlassian.net/browse/KBV-XXX)

### Other considerations

### DIFF
--- a/deploy/template.yaml
+++ b/deploy/template.yaml
@@ -48,6 +48,9 @@ Resources:
     Properties:
       CodeUri: ../dist/session.zip
       Handler: uk.gov.di.ipv.cri.address.api.handler.SessionHandler
+      Environment:
+        Variables:
+          TableName: !Sub "address-session-${AWS::StackName}"
       Policies:
         DynamoDBWritePolicy:
           TableName:
--- a/lambdas/session/src/main/java/uk/gov/di/ipv/cri/address/api/handler/SessionHandler.java
+++ b/lambdas/session/src/main/java/uk/gov/di/ipv/cri/address/api/handler/SessionHandler.java
@@ -1,5 +1,6 @@
 package uk.gov.di.ipv.cri.address.api.handler;
 
+import com.amazonaws.regions.Regions;
 import com.amazonaws.services.dynamodbv2.AmazonDynamoDB;
 import com.amazonaws.services.dynamodbv2.AmazonDynamoDBClientBuilder;
 import com.amazonaws.services.dynamodbv2.document.DynamoDB;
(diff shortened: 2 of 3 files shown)


lambdas/session/src/main/java/uk/gov/di/ipv/cri/address/api/handler/SessionHandler.java
```diff
@@ -20,6 +25,20 @@ public APIGatewayProxyResponseEvent handleRequest(
                 new APIGatewayProxyResponseEvent();
         apiGatewayProxyResponseEvent.setStatusCode(HttpStatus.SC_CREATED);
         UUID uuid = UUID.randomUUID();
+
+        try {
+            AmazonDynamoDB client = AmazonDynamoDBClientBuilder.standard().build();
```
jkunle, line 30: You want to specify the region read somewhere it helps with performance see.   https://github.com/aws-samples/serverless-patterns/blob/main/apigw-lambda-dynamodb-sam-java/src/main/java/com/example/utils/DDBUtils.java#L17

So region would be `EU_WEST_2`

I couldn't find the story related to this I suppose it is similar to our story here maybe not as elaborate https://govukverify.atlassian.net/browse/KBV-90

You could follow the `lib/persistence/dataStore` pattern

image

The ipv-core used this pattern and we've used it as well, I think it would work here as well

jkunle: https://github.com/alphagov/di-ipv-cri-kbv-api/blob/main/lib/src/main/java/uk/gov/di/ipv/cri/kbv/api/persistence/DataStore.java

deploy/template.yaml
```diff
@@ -48,6 +48,9 @@ Resources:
     Properties:
       CodeUri: ../dist/session.zip
```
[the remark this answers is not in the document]
jkunle: I suppose if there was some test, I could have approve this and the enhancement I suggested could have come letter. Since this does seen like it would write the session-id to the table

lambdas/session/src/main/java/uk/gov/di/ipv/cri/address/api/handler/SessionHandler.java
idrop: Thanks for explaining this @jkunle 👍

deploy/template.yaml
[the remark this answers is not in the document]
KaraokeStu: The zip method actually allows for two things:
1. It allows use of a shared library in the project between lambdas
2. It shrinks file sizes to be stored in deployment considerably (eg: 29MB vs 89MB)

PR #10, opened 2022-02-25
KBV-203 Store Address Session

idrop:
## Proposed changes

Save the Address CRI API session id to DynamoDB as primary index.

### What changed

* This PR builds on the `BAU-pro-structure` branch, created to amend the gradle structure of the project, and introduce a lib module.
* Save the session to a DynamoDB table `address-session-${AWS::StackName}` with a TTL of 48 hrs
* Add a `.sdkmanrc` to set the host JVM to 11 zulu, like `core-back`
* Sets SSM parameters prefixed with the current stack name. This avoids account-scoped params being managed by multiple stacks

### Why did it change

The session id needs to be stored as the identifier for calls to the Address CRI API.

### Issue tracking

- [KBV-203](https://govukverify.atlassian.net/browse/KBV-203)


MattBidewell (approved):
LGTM - its makes sense and code is easy to understand

PR #13, opened 2022-03-04
KBV-221 Validate Address CRI request JWT

idrop:
## Proposed changes

* Validate OAuth request parameters sent from Address CRI frontend to the API via the `/session` endpoint.
* Validate the JWT header, claims and signature from the `request` parameter.
* Additionally save the client id, state and redirectUri to the `address-session` table

NB.

* The SSM Params and DynamoDB table are still configured in `deploy/template.yaml` until we have a structure for deploying them in `di-ipv-config`. At that point they can be moved.
* I've used a wrapper around setting metrics and logging called `EventProbe` - look for it in `SessionHandler` - your thoughts welcome.

### Issue tracking

- [KBV-221](https://govukverify.atlassian.net/browse/KBV-221)


MattBidewell (approved):
Big one - LGTM tho

Maybe get a Java dev to give it a double look?

deploy/template.yaml
```diff
@@ -149,3 +155,31 @@ Resources:
...
+      Name: !Sub /${AWS::StackName}/clients/ipv-core-stub/jwtAuthentication/publicCertificateToVerify
+      Value: MIIDJDCCAgwCCQD3oEU83RePojANBgkqhkiG9w0BAQsFADBUMQswCQYDVQQGEwJHQjEXMBUGA1UECgwOQ2FiaW5ldCBPZmZpY2UxDDAKBgNVBAsMA0dEUzEeMBwGA1UEAwwVSVBWIENvcmUgU3R1YiBTaWduaW5nMB4XDTIyMDIwNDE3NDg1NFoXDTMyMDIwMjE3NDg1NFowVDELMAkGA1UEBhMCR0IxFzAVBgNVBAoMDkNhYmluZXQgT2ZmaWNlMQwwCgYDVQQLDANHRFMxHjAcBgNVBAMMFUlQViBDb3JlIFN0dWIgU2lnbmluZzCCASIwDQYJKoZIhvcNAQEBBQADggEPADCCAQoCggEBAMT9tGIIujF53SfiHsc+BDra/5qb/PObr8BfutWq4/9kp32eHKOBqTYberXeKJhIS80OB9AC/55QO3V2HdajJJXZbj37shvNy5HcGNxVYRFWt/qyh3+SRTbgfCnfs4QQ6uSLQkJof347qGi26kJU7RuM47grfGCahMvdEOnQgeIHLKw1yqmu8yniy0Lf48jnGlyR6r6QG7UMI2Dk5hdGOEw2WZCoSGLsXII94xS3JKB4sbjEfyuMg87o3pBjoks1LP8KXRcbkBIlc2q8DtEh2YtP57VJfdNhR/gxtHjZhL2E6q+MM158/1xwyXIGcllf+MIserihKEnmsk6wKaJcalcCAwEAATANBgkqhkiG9w0BAQsFAAOCAQEAmiAOKd9ZQSjZmIB0GBD1lJaiI1xWwHbwhhuwFsO3YzrIOmB/pDYh0e2FFdsYZ/I48VZn4WQM8mexF/3B2mnG3gXbqHtY8tRPLa0nZq53cOWczKF8RUM2xeWwitYZvfMtx6rRliCUv91IbmKHeGRGiOifOoGxvj8qp30lB2mRBuxP2N4VUAdkqWUjnTdPJr+5eHQeFTbgFg44FxJ59Mz+gGbv4/dQo8ZtFFA/Cqwvz4pevFSken4e9wRF6JGA+AuYdW3tiVutDo/UF+aFir/pDfcrCtyes9xiUv3Iu9x7zKEUG7hwMj0gG83Cvx5SuWJyPb4eKrrdRlRRmLD7PsucUw==
```
KaraokeStu, line 163: If you look at #14 you'll see I've done similar with the API Key parameter, where it is stored locally in the samdeploy.toml file - this might be a better way, alternatively, in this case since all systems SHOULD verify against the same public key, perhaps we should use a centralised parameter from the AWS parameter store, rather than developer-specific?

lambdas/session/src/main/java/uk/gov/di/ipv/cri/address/api/handler/SessionHandler.java
```diff
@@ -4,31 +4,42 @@
...
+import uk.gov.di.ipv.cri.address.library.helpers.ApiGatewayResponseGenerator;
+import uk.gov.di.ipv.cri.address.library.helpers.EventProbe;
 import uk.gov.di.ipv.cri.address.library.service.AddressSessionService;
 
 import java.util.Map;
+import java.util.UUID;
+
+import static org.apache.logging.log4j.Level.ERROR;
+import static org.apache.logging.log4j.Level.INFO;
+import static software.amazon.awssdk.http.HttpStatusCode.BAD_REQUEST;
```
KaraokeStu, line 23: Elsewhere, we're using org.apache.http.HttpStatus for status codes. We should pick one and standardise

lib/src/main/java/uk/gov/di/ipv/cri/address/library/domain/SessionRequest.java
```diff
@@ -0,0 +1,75 @@
+package uk.gov.di.ipv.cri.address.library.domain;
+
+import com.fasterxml.jackson.annotation.JsonAlias;
+
+import java.net.URI;
+import java.util.StringJoiner;
+
+public class SessionRequest {
+
+    @JsonAlias("response_type")
```
KaraokeStu, line 10: Is there a reason to use JsonAlias over JsonProperty?
JsonAlias does not work with Enums, if we wanted to use any in future

lib/src/main/java/uk/gov/di/ipv/cri/address/library/exceptions/ValidationException.java
```diff
@@ -0,0 +1,11 @@
+package uk.gov.di.ipv.cri.address.library.exceptions;
+
+public class ValidationException extends Exception {
```
KaraokeStu, line 3: Should some of these validation exceptions be specific to the class that is throwing them to make things easier to debug, or should we try to be as generic as possible?

lib/src/main/java/uk/gov/di/ipv/cri/address/library/helpers/EventProbe.java
```diff
@@ -0,0 +1,44 @@
...
+import software.amazon.lambda.powertools.metrics.MetricsUtils;
+
+import java.util.Map;
+
+public class EventProbe {
+
+    private static final Logger LOGGER = LogManager.getLogger();
+    private static final MetricsLogger metricsLogger = MetricsUtils.metricsLogger();
+
+    public EventProbe log(Level level, Exception e) {
```
KaraokeStu, line 17: Is it worth having "helper" methods such as "debug", "info", "warning" etc as per log4j?

lib/src/main/java/uk/gov/di/ipv/cri/address/library/exceptions/ValidationException.java
jkunle: I agree maybe we could do something like this

```
public GenericException extends Exception {
      public GenericException(String errorMessage, Throwable cause) {
               super(errorMessage, cause);
      }
}
```

Then we can do this

```
new GenericException(ex.getMessage(), ex.getCause());
```

lambdas/session/src/main/java/uk/gov/di/ipv/cri/address/api/handler/SessionHandler.java
jkunle: I think I prefer the `aws` so maybe we can switch to Phil use here

lib/src/main/java/uk/gov/di/ipv/cri/address/library/exceptions/ValidationException.java
KaraokeStu:
```
public class PostcodeLookupProcessingException extends RuntimeException {
    public PostcodeLookupProcessingException() {
        super();
    }

    public PostcodeLookupProcessingException(String message) {
        super(message);
    }

    public PostcodeLookupProcessingException(String message, Throwable cause) {
        super(message, cause);
    }

    public PostcodeLookupProcessingException(Throwable cause) {
        super(cause);
    }
}
```

We use exceptions like this within the address API already

lib/src/main/java/uk/gov/di/ipv/cri/address/library/helpers/EventProbe.java
idrop: I'm ok with getting this extra methods added when we need them, but for the SessionHandler the method above is all that's needed.

lib/src/main/java/uk/gov/di/ipv/cri/address/library/exceptions/ValidationException.java
idrop: I've moved these two exceptions to the `exception` package and renamed them to be more descriptive

deploy/template.yaml
idrop: This cert is specific to the `ipv-core-stub` client - and there'll be additional config for `ipv-core-dev`, `ipv-core-build` etc. The cert for the PaaS-deployed  [IPV Core Stub](https://di-ipv-core-stub.london.cloudapps.digital/) can be used for all dev stacks.

lambdas/session/src/main/java/uk/gov/di/ipv/cri/address/api/handler/SessionHandler.java
idrop: I'm ok with switching to `org.apache.http.HttpStatus` so will do this

lib/src/main/java/uk/gov/di/ipv/cri/address/library/domain/SessionRequest.java
idrop: This allows us to alias the request params from the [OAuth authorization code grant request](https://datatracker.ietf.org/doc/html/rfc6749#section-4.1.3) i.e. we receive `response_type` as a request param to our frontend, which maps these params to a json payload to the /session API.

KaraokeStu (approved):
All looks good to me

PR #15, opened 2022-03-11
Kbv-146 address cri api token resource

jkunle:
## Proposed changes

A token endpoint specified as an API gateway resource that conforms to the RFC defined for  [RFC 6749 - The OAuth 2.0 Authorization Framework](https://datatracker.ietf.org/doc/html/rfc6749#section-4.1.3)
A lambda function providing an access token as defined in [RFC 6749 - The OAuth 2.0 Authorization Framework](https://datatracker.ietf.org/doc/html/rfc6749#section-4.1.4) .

The refresh token and token_type don’t need to be provided at the moment
The logic is:

validate the request per [RFC 6749 - The OAuth 2.0 Authorization Framework](https://datatracker.ietf.org/doc/html/rfc6749#section-4.1.3) 

Provide a token for the given authorization code, as a UUID-4. 
Since there is currently no **authorization code** a UUID is currently stored in the dynamodb when the session endpoint
hit. And a session id is created

The logic is very similar to with the Passport CRI token request logic, see [AccessTokenHandler.java](https://github.com/alphagov/di-ipv-cri-uk-passport-back/blob/main/lambdas/accesstoken/src/main/java/uk/gov/di/ipv/cri/passport/accesstoken/AccessTokenHandler.java)
The only difference is the all persistence is done in the AddressTable

### What changed

A new Api gate way and associated lambda resource added 

### Why did it change

We want to be able to generate an accesstoken when requesting this endpoint by supplying an existing authorization code

In API gateway go to the /session endpoint 

```
{
  "response_type": "code",
  "client_id": "ipv-core",
  "state": "state",
  "redirect_uri": "https://www.example/com/callback",
  "request": "eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJzdWIiOiJ0dCIsImlhdCI6MTUxNjIzOTAyMn0.SbcN-ywpLObhMbbaMCtW1Un8LYhQzHsEth9LvTk4oQQ"
}
```

Go to dynamodb and locate the **authorization_code** value and substitute
in the body of the request to the `/token`

```
code=authorization_code&redirect_uri=https://www.example/com/callback&grant_type=authorization_code&client_id=ipv-core
```


lambdas/accesstoken/src/main/java/uk/gov/di/ipv/cri/address/api/handler/AccessTokenHandler.java
```diff
@@ -0,0 +1,117 @@
...
+                String message =
+                        String.format(
+                                "Session with id:%s could not be found in the database.",
+                                sessionId);
+                LOGGER.error(message);
+                return ApiGatewayResponseGenerator.proxyJsonResponse(
+                        OAuth2Error.INVALID_GRANT.getHTTPStatusCode(),
+                        OAuth2Error.INVALID_GRANT.toJSONObject());
+            }
+            if (addressSessionItem.getAuthorizationCode() == authorizationCodeFromRequest) {
```
amanm-alphagov, line 77: use `.equalsIgnoreCase` for String comparison

lambdas/accesstoken/build.gradle
```diff
@@ -0,0 +1,30 @@
...
+			configurations.gson,
+			configurations.jackson
+
+	aspect configurations.powertools
+
+	testImplementation configurations.tests
+	testRuntimeOnly configurations.test_runtime
+}
+
+task buildZip(type: Zip) {
```
amanm-alphagov, line 19: Do we need this task because we dont use it in the deployment [here](https://github.com/alphagov/di-ipv-cri-address-api/blob/4969c03c3c29ea9300fbce9a20ebbd5c5f28a6e9/.github/workflows/deploy.yml#L44)

lambdas/accesstoken/src/main/java/uk/gov/di/ipv/cri/address/api/handler/AccessTokenHandler.java
```diff
@@ -0,0 +1,117 @@
...
+import com.nimbusds.oauth2.sdk.AuthorizationCodeGrant;
+import com.nimbusds.oauth2.sdk.ErrorObject;
+import com.nimbusds.oauth2.sdk.OAuth2Error;
+import com.nimbusds.oauth2.sdk.ParseException;
+import com.nimbusds.oauth2.sdk.TokenRequest;
+import com.nimbusds.oauth2.sdk.TokenResponse;
+import com.nimbusds.oauth2.sdk.http.HTTPRequest;
+import org.apache.http.HttpStatus;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
```
amanm-alphagov, line 18: Are we using loggers from [apache](https://github.com/alphagov/di-ipv-cri-address-api/blob/4969c03c3c29ea9300fbce9a20ebbd5c5f28a6e9/lambdas/session/src/main/java/uk/gov/di/ipv/cri/address/api/handler/SessionHandler.java#L9) framework or the `slf4j`

build.gradle
```diff
@@ -56,6 +56,10 @@ subprojects {
 	}
 
 	dependencies {
+		jackson "com.fasterxml.jackson.core:jackson-core:${dependencyVersions.jackson_version}",
```
idrop, line 59: there are two `jackson` blocks now, could you merge please?

deploy/api.yaml
```diff
@@ -82,8 +82,132 @@ paths:
...
+        responses:
+          default:
+            statusCode: "200"
+        passthroughBehavior: "when_no_match"
+        contentHandling: "CONVERT_TO_TEXT"
+        type: "aws_proxy"
 components:
+  parameters:
+    SessionHeader:
+      name: session_id
```
idrop, line 133: IPV Core won't be sending the `session_id` in this OAuth flow call. Rather, we use the authorization `code` to look up the session id

lambdas/accesstoken/build.gradle
idrop: the `buildZip` task is used in the block below, see `build.finalizedBy(buildZip)`

lib/src/main/java/uk/gov/di/ipv/cri/address/library/helpers/RequestHelper.java
```diff
@@ -25,4 +30,14 @@ public static String getHeaderByKey(Map<String, String> headers, String headerKe
         }
         return null;
     }
+
+    public static Map<String, String> parseRequestBody(String body) {
```
idrop, line 34: Don't think this function is being used

deploy/api.yaml
```diff
@@ -82,8 +82,124 @@ paths:
...
+        - "code"
+        - "redirect_uri"
+        - "client_id"
+      type: "object"
+      properties:
+        grant_type:
+          type: "string"
+          pattern: "authorization_code"
+          example: "authorization_code"
+        code:
```
idrop, line 172: If there's no minLength on the code does it mean API gateway will accept an empty value i.e.`...&code=&...`

deploy/template.yaml
```diff
@@ -137,6 +136,44 @@ Resources:
...
+            - Effect: Allow
+              Action:
+                - ssm:GetParameter
+              Resource:
+                - !Sub "arn:aws:ssm:${AWS::Region}:${AWS::AccountId}:parameter/${AWS::StackName}/AddressSessionTableName"
+                - !Sub "arn:aws:ssm:${AWS::Region}:${AWS::AccountId}:parameter/${AWS::StackName}/AddressSessionTtl"
+                - !Sub "arn:aws:ssm:${AWS::Region}:${AWS::AccountId}:parameter/${AWS::StackName}/clients/*"
+
+
+  AddressCriCodeSigning:
```
idrop, line 168: I don't think we need this AWS::Lambda::CodeSigningConfig resource per lambda anymore. @mrdoodles can you confirm?

lib/src/main/java/uk/gov/di/ipv/cri/address/library/service/AddressSessionService.java
```diff
@@ -60,6 +79,13 @@ public AddressSessionService(
         this.clock = clock;
     }
 
+    public ValidationResult<ErrorObject> validateTokenRequest(TokenRequest tokenRequest) {
+        if (!tokenRequest.getAuthorizationGrant().getType().equals(GrantType.AUTHORIZATION_CODE)) {
```
idrop, line 83: the api gateway is already doing this validation - personally I think we should have integration tests that check this rather than a unit test

lambdas/accesstoken/src/main/java/uk/gov/di/ipv/cri/address/api/handler/AccessTokenHandler.java
```diff
@@ -0,0 +1,115 @@
...
+    }
+
+    public AccessTokenHandler() {
+        this.addressSessionService = new AddressSessionService();
+    }
+
+    public APIGatewayProxyResponseEvent handleRequest(
+            APIGatewayProxyRequestEvent input, Context context) {
+        try {
+            TokenRequest tokenRequest = createTokenRequest(input.getBody());
```
idrop, line 43: Could we move all validation into AddressSessionService, and have that method return a TokenRequest?

lambdas/accesstoken/src/main/java/uk/gov/di/ipv/cri/address/api/handler/AccessTokenHandler.java
```diff
@@ -0,0 +1,115 @@
...
+                        OAuth2Error.INVALID_GRANT.toJSONObject());
+            }
+            if (!authorizationCodeFromRequest.equals(addressSessionItem.getAuthorizationCode())) {
+                LOGGER.error(
+                        "Access Token could not be issued. The supplied authorization code was not found in the database.");
+                return ApiGatewayResponseGenerator.proxyJsonResponse(
+                        OAuth2Error.INVALID_GRANT.getHTTPStatusCode(),
+                        OAuth2Error.INVALID_GRANT.toJSONObject());
+            }
+
```
idrop, line 79: We should validate the request_uri is identical to that supplied in the authorization request. We store this value. See the redirect_uri validation requirements at https://datatracker.ietf.org/doc/html/rfc6749#section-4.1.3

lib/build.gradle
```diff
@@ -5,6 +5,7 @@ plugins {
 dependencies {
 
 	implementation configurations.lambda,
+			configurations.jackson,
```
idrop, line 8: this dep is already included

deploy/template.yaml
```diff
@@ -20,12 +20,22 @@ Parameters:
...
+    Type: String
+    Default: testapikey
+
   OrdnanceSurveyAPIURL:
     Description: Ordnance Survey API URL
     Type: String
     Default: "https://api.os.uk/search/places/v1/postcode"
 
 Conditions:
+  UseCodeSigning:
```
idrop, line 34: I don't see this being used anywhere, can we delete?

deploy/template.yaml
```diff
@@ -20,6 +20,11 @@ Parameters:
       - prod
     ConstraintDescription: specify dev, build, staging, integration or prod for environment
 
+  OrdnanceSurveyAPIKey:
```
idrop, line 23: this is not present in main

deploy/template.yaml
```diff
@@ -243,19 +274,29 @@ Resources:
       Name: !Sub /${AWS::StackName}/clients/ipv-core-stub/jwtAuthentication/issuer
       Value: ipv-core-stub
 
+  SecretOrdnanceSurveyAPIKey:
```
idrop, line 277: this is not present in main

idrop (approved):
Thanks for this Ola

PR #27, opened 2022-03-18
[KBV-268] add address api parameters deployments

mrdoodles:
### What changed

* Parameters added to separate infra stack under /infra directory
* Existing automated GitHub workflows updated to ignore changes under the /infra directory 
* New validation and deployment workflows added to deploy /infra and to only run on /infra changes

### Why did it change

So that we can deploy supporting parameters and possibly other supporting infrastructure independently of the api deployments.

### Issue tracking

- [KBV-268](https://govukverify.atlassian.net/browse/KBV-268)

- [X] Documented in the [README](./blob/main/README.md)

### Other considerations

The infra deployment requires an additional GitHub environment to be configured with parameters for the second stack.
- [X] Updated [README](./blob/main/README.md)


.github/workflows/deploy-infra.yml
```diff
@@ -0,0 +1,48 @@
...
+        uses: aws-actions/setup-sam@v1
+
+      - name: Assume temporary AWS role
+        uses: aws-actions/configure-aws-credentials@v1
+        with:
+          role-to-assume: ${{ secrets.GH_ACTIONS_ROLE_ARN }}
+          aws-region: ${{ env.AWS_REGION }}
+      
+      - name: SAM build
+        run: sam build -t infra/template.yaml --config-env ${{ env.ENVIRONMENT }}
```
idrop, line 33: I think we can drop the `--config-env` flag for this `sam build` and the `sam package` below.

.github/workflows/build.yml
```diff
@@ -2,6 +2,8 @@ name: Build
 on:
   pull_request:
     types: [opened, synchronize, reopened]
+    paths-ignore:
+      - 'infra/**'
```
idrop, line 6: I'd like all our deployed resources to be under one root dir, does it make sense to have a structure like:
````
infrastructure/
├── lambda/
├── parameter/
````
We can defer this to another clean-up PR though if it muddies this PR.

File: .github/workflows/validate-infra.yml [hunk 3]
```diff
@@ -0,0 +1,32 @@
...
+        uses: aws-actions/setup-sam@v1
+
+      - name: Assume temporary AWS role
+        uses: aws-actions/configure-aws-credentials@v1
+        with:
+          role-to-assume: ${{ secrets.GH_ACTIONS_ROLE_ARN }}
+          aws-region: ${{ env.AWS_REGION }}
+
+      - name: SAM Validate
+        run: sam validate --region ${{ env.AWS_REGION }} -t deploy/template.yaml --config-env ${{ env.ENVIRONMENT }}
```
idrop, line 32: we don't need config-env here

.github/workflows/build.yml
mrdoodles: Infrastructure/lambda, would this be both the sam template and the api yaml?

PR #28, opened 2022-03-19
KBV-274 Deploy a pre-merge integration test stack to dev

idrop:
## Proposed changes

* Pre-merge checks as separate jobs, with inspiration from the https://github.com/alphagov/di-authentication-api github actions
* Pre-merge checks now show correct code coverage in the sonarcloud check
* Pre-merge deployment of an ephemeral stack to dev, after pre-merge checks run. The stack is always torn down.

### How to test

* Observe that after pre-merge checks pass, an ephemeral stack of the API is deployed to the aws dev account

### Why did it change

Support integration testing in dev


mrdoodles (approved):

.github/workflows/pre-merge-integration-test.yml
```diff
@@ -0,0 +1,64 @@
...
+      - name: Set short SHA
+        id: vars
+        run: echo "::set-output name=sha_short::$(git rev-parse --short HEAD)"
+
+      - name: SAM deploy integration test stack
+        run: |
+          STACK_NAME=${{ env.STACK_NAME }}-${{ steps.vars.outputs.sha_short }}
+          sam deploy -t deploy/template.yaml \
+            --no-fail-on-empty-changeset --no-confirm-changeset ${{ env.ROLLBACK_ACTION }} \
+            --signing-profiles SessionFunction=${{ secrets.SIGNING_PROFILE_NAME }} \
```
subnova, line 55: Only post-merge code should be signed as the point about signing is that it proves that the application was built by the pipeline from reviewed code. If you apply the latest version of the pipeline stack it will prevent you from using the associated signing key in pre-merge scenarios: see PLAT-62

.github/workflows/pre-merge-integration-test.yml
```diff
@@ -0,0 +1,64 @@
...
+        with:
+          java-version: 11
+
+      - name: Setup SAM
+        uses: aws-actions/setup-sam@v1
+
+      - name: Assume temporary AWS role
+        uses: aws-actions/configure-aws-credentials@v1
+        with:
+          role-to-assume: ${{ secrets.GH_ACTIONS_ROLE_ARN }}
```
subnova, line 37: We should not share accounts for code that is part of the production pipeline and code that has yet to be reviewed because of the risk of a vulnerability being created that is then used to corrupt the production pipeline. i.e. build and "dev" should be different accounts.

idrop: I think they are different accounts, as their [github environment ](https://docs.github.com/en/actions/deployment/targeting-different-environments/using-environments-for-deployment)is different -  @mrdoodles can you confirm?

.github/workflows/pre-merge-integration-test.yml
idrop: 👍 will remove for the dev environment

.github/workflows/pre-merge-checks.yml
```diff
@@ -0,0 +1,124 @@
...
+          path: |
+            .gradle/
+            */build/
+            !*/build/reports
+            !*/build/jacoco
+          key: ${{ runner.os }}-build-${{ github.sha }}
+      - name: Cache Unit Test Reports
+        uses: actions/cache@v2
+        with:
+          key: ${{ runner.os }}-unit-test-reports-${{ github.sha }}
```
merlinc, line 115: If the commit hasn't changed, don't rerun the reports 👍

.github/workflows/pre-merge-integration-test.yml
```diff
@@ -0,0 +1,70 @@
...
+            --s3-bucket ${{ secrets.ARTIFACT_SOURCE_BUCKET_NAME }} \
+            --s3-prefix $STACK_NAME \
+            --region ${{ env.AWS_REGION }} \
+            --capabilities CAPABILITY_IAM
+
+      - name: Run API integration tests
+        run: echo "🤞 now run integration tests..."
+
+      - name: Delete integration test stack
+        if: always()
```
merlinc, line 67: Do we need this line?

idrop: we need it to always tear down the stack

mrdoodles (approved):
Looks OK to me but won't deploy to build until I have re-created the CF infrastructure

PR #32, opened 2022-03-22
[KBV-237] Store passed addresses in DynamoDB and return an Authentication Result

KaraokeStu:
### What changed

Add the ability to save one or more addresses to the Address CRI Session table in DynamoDB and return an authorization code back to the user if we have one or more addresses.

### Why did it change

The Address CRI needed to be able to save one or more addresses back to the session ready to confirm we have details ready to be passed to DBS

### Issue tracking

- [KBV-237](https://govukverify.atlassian.net/browse/KBV-237)

- [x] No environment variables or secrets were added or changed


deploy/api.yaml
```diff
@@ -4,6 +4,60 @@ info:
...
+            application/json:
+              schema:
+                $ref: "#/components/schemas/Error"
+        500:
+          description: "Internal server error"
+          content:
+            application/json:
+              schema:
+                $ref: "#/components/schemas/Error"
+      x-amazon-apigateway-request-validator: "Validate body"
```
idrop, line 49: can we use "Validate both" so we get validation on the session id in the header

deploy/template.yaml
```diff
@@ -25,7 +25,19 @@ Parameters:
...
     Default: "https://api.os.uk/search/places/v1/postcode"
 
+  OrdnanceSurveyAPIKey:
+    Description: Ordnance Survey API Key
+    Type: String
+    Default: testapikey
+
+
 Conditions:
+  UseCodeSigning:
```
idrop, line 35: This condition is evaluated in the globals block below, so you shouldn't have to apply it to your function

deploy/template.yaml
```diff
@@ -140,6 +156,52 @@ Resources:
...
+            - Effect: Allow
+              Action:
+                - ssm:GetParametersByPath
+              Resource:
+                - !Sub "arn:aws:ssm:${AWS::Region}:${AWS::AccountId}:parameter/${AWS::StackName}/clients/*"
+
+  AddressFunction:
+    Type: AWS::Serverless::Function
+    Properties:
+      CodeSigningConfigArn: !If
```
idrop, line 176: this could be removed as its handled in the globals block

deploy/template.yaml
```diff
@@ -84,6 +96,7 @@ Resources:
       OpenApiVersion: 3.0.1
       EndpointConfiguration:
         Type: REGIONAL
+    DependsOn: [SessionFunction, PostcodeLookupFunction, AddressFunction, AccessTokenFunction]
```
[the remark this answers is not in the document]
KaraokeStu: While not strictly necessary, it does ensure the template processes in the correct order and ensures all these functions create properly before it tries to create the api gateway - it SHOULD work it out by itself, so I cam remove if required

deploy/template.yaml
```diff
@@ -278,18 +330,31 @@ Resources:
 
   SessionFunctionPermission:
     Type: AWS::Lambda::Permission
+    DependsOn:
```
idrop, line 333: these permissions seem to be working on main without this DependsOn, so can we remove?

deploy/template.yaml
```diff
@@ -278,18 +330,31 @@ Resources:
...
+      - SessionFunction
     Properties:
       Action: lambda:InvokeFunction
       FunctionName: !GetAtt SessionFunction.Arn
       Principal: apigateway.amazonaws.com
 
   PostcodeLookupFunctionPermission:
     Type: AWS::Lambda::Permission
+    DependsOn:
```
idrop, line 342: these permissions seem to be working on main without this DependsOn, so can we remove?

lambdas/postcode-lookup/src/main/java/uk/gov/di/ipv/cri/address/api/handler/PostcodeLookupHandler.java
```diff
@@ -41,24 +55,37 @@ public PostcodeLookupHandler() {
     public APIGatewayProxyResponseEvent handleRequest(
             APIGatewayProxyRequestEvent input, Context context) {
 
+        String sessionId = input.getHeaders().get(SESSION_ID);
+        eventProbe.addDimensions(Map.of(SESSION_ID, sessionId));
```
idrop, line 59: These dimensions allows us to partition our metrics in CloudWatch. I don't think we need to add any dimensions to the eventProbe for this lambda. But I think we are missing pushing a counter metric for the happy case i.e. `eventProbe.counterMetric("postcode_lookup");` and also pushing that same metric for all exception cases i.e. `eventProbe.log(INFO, e).counterMetric("postcode_lookup", 0d);`

This will allow us to dashboard/alert on the success rate of this lambda

KaraokeStu: Resolved with latest changes

lib/src/main/java/uk/gov/di/ipv/cri/address/library/models/AuthorizationResponse.java
```diff
@@ -0,0 +1,30 @@
...
+        this.code = session.getAuthorizationCode();
+        this.redirect_uri = session.getRedirectUri();
+        this.state = session.getState();
+    }
+
+    public UUID getCode() {
+        return code;
+    }
+
+    public URI getRedirect_uri() {
```
jkunle, line 23: getRedirectUri

lib/src/main/java/uk/gov/di/ipv/cri/address/library/models/CanonicalAddress.java
```diff
@@ -1,27 +1,36 @@
...
-    private final String buildingNumber;
-    private final String dependentThoroughfare;
-    private final String doubleDependentLocality;
-    private final String dependentLocality;
-    private final String buildingName;
-    private final String thoroughfareName;
-    private final String postTown;
-    private final String postcode;
-    private final String countryCode;
+    private Optional<Integer> uprn;
```
jkunle, line 13: I'm not sure this is idiomatic use of optional, I mean declaring fields as Optional rather I think the returning method could be optional instead, I don't know how this would affect the stuff you've done so maybe something like this is more idiomatic 

private Integer uprn;

 public Optional<Integer> getUprn() {
      return Optional.ofNullable(uprn);
 }

 public void setUprn(Integer uprn) {
      this.uprn = uprn;
  }

lib/src/main/java/uk/gov/di/ipv/cri/address/library/persistence/item/AddressSessionItem.java
```diff
@@ -1,10 +1,12 @@
 package uk.gov.di.ipv.cri.address.library.persistence.item;
 
-import software.amazon.awssdk.enhanced.dynamodb.mapper.annotations.DynamoDbBean;
-import software.amazon.awssdk.enhanced.dynamodb.mapper.annotations.DynamoDbPartitionKey;
-import software.amazon.awssdk.enhanced.dynamodb.mapper.annotations.DynamoDbSecondaryPartitionKey;
+import software.amazon.awssdk.enhanced.dynamodb.mapper.annotations.*;
```
jkunle, line 3: Please can you make this import explicit

lib/src/main/java/uk/gov/di/ipv/cri/address/library/persistence/item/AddressSessionItem.java
```diff
@@ -15,13 +17,15 @@
     private String clientId;
     private String state;
     private URI redirectUri;
+    private List<CanonicalAddressWithResidency> addresses;
```
jkunle, line 20: Why not 

`private List<CanonicalAddressWithResidency> addresses = new ArrayList<>();`

lambdas/address/src/main/java/uk/gov/di/ipv/cri/address/api/handler/AddressHandler.java
```diff
@@ -0,0 +1,78 @@
...
+
+    public AddressHandler(AddressSessionService sessionService, EventProbe eventProbe) {
+        this.sessionService = sessionService;
+        this.eventProbe = eventProbe;
+    }
+
+    @Override
+    @Logging(correlationIdPath = CorrelationIdPathConstants.API_GATEWAY_REST)
+    @Metrics(captureColdStart = true)
+    public APIGatewayProxyResponseEvent handleRequest(
```
idrop, line 47: I'd like to see test coverage on this method please - at least some assertions about which exceptions cause what http status codes. Also some assertions on what we're sending to the eventProbe.

lib/src/main/java/uk/gov/di/ipv/cri/address/library/service/AddressSessionService.java
```diff
@@ -294,6 +289,67 @@ private boolean validSignature(SignedJWT signedJWT, PublicKey clientPublicKey)
         }
     }
 
+    public List<CanonicalAddressWithResidency> parseAddresses(String addressBody)
+            throws AddressProcessingException {
+        List<CanonicalAddressWithResidency> addresses;
```
idrop, line 294: I'd like to see a test that checks that we can marshall a String to a List<CanonicalAddressWithResidency>. We could test this with integration tests but I think there's value in checking jackson is doing its job. We might want to check how the Optional<> setters work

lib/src/main/java/uk/gov/di/ipv/cri/address/library/service/AddressSessionService.java
```diff
@@ -294,6 +289,67 @@ private boolean validSignature(SignedJWT signedJWT, PublicKey clientPublicKey)
...
+        return dataStore.getItem(sessionId);
+    }
+
+    public void validateSessionId(String sessionId)
+            throws SessionValidationException, SessionNotFoundException, SessionExpiredException {
+        if (sessionId == null) {
+            throw new SessionValidationException("session id is null");
+        }
+        if (sessionId.isEmpty()) {
+            throw new SessionValidationException("session id is empty");
```
idrop, line 316: I'm conflicted on whether we should even have these guards - as we're meant to be validating for a uuid on the api gateway. @jkunle I've see that you also re-check some stuff that API gateway is checking - we should all have a chat about whether we need to do this more than once.

lib/src/main/java/uk/gov/di/ipv/cri/address/library/service/AddressSessionService.java
```diff
@@ -294,6 +289,67 @@ private boolean validSignature(SignedJWT signedJWT, PublicKey clientPublicKey)
...
+        }
+
+        AddressSessionItem sessionItem = dataStore.getItem(sessionId);
+        if (sessionItem == null) {
+            throw new SessionNotFoundException("session not found");
+        }
+
+        if (sessionItem.getExpiryDate() < clock.instant().getEpochSecond()) {
```
idrop, line 324: I'd like this logic to have some test coverage

lib/src/main/java/uk/gov/di/ipv/cri/address/library/service/AddressSessionService.java
```diff
@@ -294,6 +289,67 @@ private boolean validSignature(SignedJWT signedJWT, PublicKey clientPublicKey)
...
+            throw new SessionExpiredException("session expired");
+        }
+    }
+
+    public AddressSessionItem saveAddresses(
+            String sessionId, List<CanonicalAddressWithResidency> addresses)
+            throws SessionExpiredException, SessionValidationException, SessionNotFoundException {
+        validateSessionId(sessionId);
+
+        var sessionItem = dataStore.getItem(sessionId);
```
idrop, line 334: in the method above we're doing the same getItem call, but follow with a null check. Is there a reason why there isn't that check here?

lib/src/main/java/uk/gov/di/ipv/cri/address/library/service/AddressSessionService.java
```diff
@@ -294,6 +289,67 @@ private boolean validSignature(SignedJWT signedJWT, PublicKey clientPublicKey)
...
+        sessionItem.setAddresses(addresses);
+        if (!addresses.isEmpty()) {
```
idrop, line 336: we've already checked addresses are not empty higher up the stack

lib/src/main/java/uk/gov/di/ipv/cri/address/library/service/AddressSessionService.java
```diff
@@ -294,6 +289,67 @@ private boolean validSignature(SignedJWT signedJWT, PublicKey clientPublicKey)
...
+            // We can create an authorization code
+            sessionItem.setAuthorizationCode(UUID.randomUUID());
+        }
+        dataStore.update(sessionItem);
+
+        return sessionItem;
+    }
+
+    public List<CanonicalAddressWithResidency> getAddresses(String sessionId)
```
idrop, line 345: my IDE is telling me this method isn't used

idrop (dismissed):
Thanks Stu

PR #33, opened 2022-03-22
[KBV-268] address api parameters

mrdoodles:
### What changed

* Parameters added to separate infra stack under /infrastructure/parameter directory
* Existing automated GitHub workflows updated to watch specific paths 
* New validation and deployment workflows added to watch and deploy /infrastructure/parameter changes
* SAM template and API Gateway configuration moved under /infrastructure/lambda

### Why did it change

* So that we can deploy supporting parameters and possibly other supporting infrastructure independently of the api deployments.
* For cleaner separation of infrastructure with differing lifecycles.

### Issue tracking

- [KBV-268](https://govukverify.atlassian.net/browse/KBV-268)

- [X] Documented in the [README](./blob/main/README.md)

### Other considerations

The infra deployment requires an additional GitHub environment to be configured with parameters for the second stack.
- [X] Updated [README](./blob/main/README.md)


.github/workflows/build.yml
```diff
@@ -2,6 +2,14 @@ name: Build
 on:
   pull_request:
     types: [opened, synchronize, reopened]
+    paths:
```
idrop, line 5: is there no way we can do an ignore paths instead?

.github/workflows/deploy-infra.yml
```diff
@@ -0,0 +1,51 @@
...
+      - name: Setup SAM
+        uses: aws-actions/setup-sam@v1
+
+      - name: Assume temporary AWS role
+        uses: aws-actions/configure-aws-credentials@v1
+        with:
+          role-to-assume: ${{ secrets.PARAM_GH_ACTIONS_ROLE_ARN }}
+          aws-region: ${{ env.AWS_REGION }}
+
+      - name: SAM Validate
```
idrop, line 32: Do we need to do another sam validate here?

.github/workflows/post-merge-deploy-to-dev.yml
```diff
@@ -38,17 +46,29 @@ jobs:
           aws-region: ${{ env.AWS_REGION }}
 
       - name: SAM Validate
+<<<<<<< HEAD
```
idrop, line 49: this looks like a merge conflict

infrastructure/parameter/template.yaml
```diff
@@ -0,0 +1,117 @@
...
+
+  IPVCoreStubIssuerMapping:
+    Environment:
+      dev: "ipv-core-stub"
+      build: "ipv-core-stub"
+      staging: "ipv-core-stub"
+      integration: "ipv-core-stub"
+      production: "ipv-core-stub"
+
+Resources:
```
idrop, line 74: do these params still exist in the infrastructure/lambda/template.yaml file

.github/workflows/build.yml
mrdoodles: We can, this is how I had it before but we end up with a similar situation needing to ignore some individual files and paths i.e. README.md, deploy.sh to name just a couple. I can reverse it if you want?

.github/workflows/deploy-infra.yml
mrdoodles: We don't have another at present, it is disabled: The PR validation of the SAM infrastructure cannot run because it requires to assume the AWS role, however it is not running on the main branch so correctly is denied access to the environment secrets. We need an offline validator for this, I have disabled the validation workflow for now and added the validation into the deployment here for now.

.github/workflows/build.yml
idrop: can we do something like
- '**'
- '!infrastructure/parameter/**'

infrastructure/lambda/template.yaml
```diff
@@ -213,53 +241,49 @@ Resources:
       Type: String
       Description: address session dynamodb table name
 
-  ParameterOrdnanceSurveyAPIURL:
+  AddressSessionTtlParameter:
     Type: AWS::SSM::Parameter
     Properties:
-      Name: !Sub "/${AWS::StackName}/OrdnanceSurveyAPIURL"
-      Value: !Ref OrdnanceSurveyAPIURL
+      Name: !Sub /${EnvironmentName}/AddressSessionTtl
```
idrop, line 247: I'd prefer to keep these as prefixed with `${AWS::StackName}`

PR #34, opened 2022-03-22
KBV-227 Protect Address CRI API with an api key

idrop:
## Proposed changes

Protect all resources on the Address CRI API by requiring clients to use an API key. These clients are the Address CRI frontend (currently in PaaS) and the IPV Core (cross-account in AWS).

This PR lands the UsagePlan and creates the keys, but does not require clients to use the keys yet.

Clients will have to supply the api key in the http header `x-api-key`.

The API key values can be found in the AWS Console.

This has been tested on a stack in our cri dev environment.

### What changed

* Add two api keys for the address cri api gateway - we provision two so we can share the 2nd key if/when required
* Add an API usage plan for our gateway with burst and rate limits
* Add a commented out usage of that api key in our OpenApi definition. This protects the entire API gateway, but its possible to set this security per resource in OpenApi.

- [KBV-227](https://govukverify.atlassian.net/browse/KBV-227)

### DIFF
--- a/infrastructure/lambda/api.yaml
+++ b/infrastructure/lambda/api.yaml
@@ -393,7 +393,14 @@ components:
         currentResidency:
           type: "boolean"
           example: true
-
+#  securitySchemes:
+#    api_key:
+#      type: "apiKey"
+#      name: "x-api-key"
+#      in: "header"
+#
+#security:
+#  - api_key: []
 
 x-amazon-apigateway-request-validators:
   Validate both:
--- a/infrastructure/lambda/template.yaml
+++ b/infrastructure/lambda/template.yaml
@@ -286,6 +286,47 @@ Resources:
         AttributeName: expiry-date
         Enabled: true
 
+  ApiUsagePlan:
+    Type: AWS::ApiGateway::UsagePlan
+    DependsOn:
+      - AddressApiStage
+    Properties:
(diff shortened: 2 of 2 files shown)


KaraokeStu (dismissed):
LGTM

deploy/api.yaml
```diff
@@ -291,6 +291,14 @@ components:
         countryCode:
           type: "string"
           example: "GBR"
+#  securitySchemes:
```
MattBidewell, line 294: i assume when we want them to be required we uncomment this code block?

MattBidewell (approved):

deploy/api.yaml
idrop: Yep that's the idea Matt 👍

infrastructure/lambda/api.yaml
```diff
@@ -393,7 +393,14 @@ components:
...
           example: true
-
+#  securitySchemes:
+#    api_key:
+#      type: "apiKey"
+#      name: "x-api-key"
+#      in: "header"
+#
+#security:
+#  - api_key: []
```
idrop, line 404: ☝️ this is commented out until clients start sending their api keys in the http header. Note that this security covers all resources on this API

MattBidewell (approved):

PR #35, opened 2022-03-24
BAU: Fix Non key value retrieval for redirectUri

jkunle:
## Why

When the item is retrieved through the Global Index using the authorizationCode, the `redirectUri` field is blank. This results in a Null pointer Exception

### What changed

Added `redirectUri` as a Non Key Projection to the authorizationCode-index global index

### Other considerations

- The AddressSessionService requires more test coverage this would be added as a subsequent PR

### DIFF
--- a/lib/src/main/java/uk/gov/di/ipv/cri/address/library/service/AddressSessionService.java
+++ b/lib/src/main/java/uk/gov/di/ipv/cri/address/library/service/AddressSessionService.java
@@ -21,6 +21,9 @@ import com.nimbusds.oauth2.sdk.http.HTTPRequest;
 import com.nimbusds.oauth2.sdk.token.AccessToken;
 import com.nimbusds.oauth2.sdk.token.BearerAccessToken;
 import com.nimbusds.oauth2.sdk.token.Tokens;
+import software.amazon.awssdk.core.pagination.sync.SdkIterable;
+import software.amazon.awssdk.enhanced.dynamodb.DynamoDbIndex;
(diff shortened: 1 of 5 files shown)


lib/src/main/java/uk/gov/di/ipv/cri/address/library/service/AddressSessionService.java
```diff
@@ -138,7 +142,7 @@ public TokenRequest createTokenRequest(String requestBody)
         boolean invalidTokenRequest =
                 request.getQueryParameters()
                         .keySet()
-                        .containsAll(List.of(CODE, CLIENT_ID, REDIRECT_URI, GRANT_TYPE));
+                        .containsAll(Set.of(CODE, CLIENT_ID, REDIRECT_URI, GRANT_TYPE));
```
idrop, line 145: the api gateway should be doing this validation - and I'm torn about whether we should chek it twice - not a blocker, more an FYI that we need a conversation about this

jkunle: Ok I understand, I have done this b'cos of Will Palmer, I think this issue came up in one of our team wide discussion and he seemed to suggest that have multiple validation doesn't hurt. I suppose if we to bypass api-gateway for any reason then it might be useful. Can we call api gateway endpoints directly in a statemachine?

idrop (approved):
We can hit the /accesstoken resource when we deploy this stack, so LGTM

PR #36, opened 2022-03-24
BAU: Increasing the coverage for access token operations

jkunle:
## Proposed changes

Added more test coverage to handle access token operation

### What changed

A major refactoring on KBV-146 was to move validation logic from the handler into the AddressSessionService
However, some of the methods in the service do not have direct test coverage this PR add more test coverage

### Other considerations

- Still not a 100% but sufficient enough


idrop (approved):
Thanks LGTM

PR #37, opened 2022-03-24
KBV-142 Save subject from jwt claim to dynamo

idrop:
## Proposed changes

Save the JWT claim subject to the session-address table so we can send it back in the verifiable credential.

This has been tested on a stack in the dev account, so that the subject appears as a new field.

### Issue tracking

- [KBV-142](https://govukverify.atlassian.net/browse/KBV-142)


idrop:
we're only getting the sub on the /credential/issue request at the end of the OAuth flow

PR #52, opened 2022-03-31
KBV-156: credential issue retrieve addresses

jkunle:
## Proposed changes

Implement the request handling for the /credential/issue resource for the Address API. This API is expected to be called by the IPV Core backend directly as the final part of the OAuth flow.

This API will accept a token and subject, and respond with a json payload of the addresses.

:info: :warning: this story does not send the VC back as a response - that will be handled in another task.

### What changed

A new endpoint specified in`OpenAPI` to issue credentials called `/credential/issue`

Clients send bearer tokens passed in the Authorization header defined by [RFC6750](https://datatracker.ietf.org/doc/html/rfc6750).

```
     POST /credential/issue HTTP/1.1
     Content-Type: application/x-www-form-urlencoded
     x-api-key: <an api key>
     Authorization: Bearer <access token>

     sub=<subject supplied by IPV Core>
```

The API should provide as response:

200 status code, and an array of addresses, or
400 if there is no item for the token
500 on an unexpected dynamodb error

### Issue tracking

- [KBV-156](https://govukverify.atlassian.net/browse/KBV-156)

### Other considerations

- [ ] Note the API is not currently using api key


idrop:
Hey Ola, based on some very recent decisions, we no longer need to send the `sub` in the request. So the API can be simplified to `GET` like:
````
     GET /credential/issue HTTP/1.1
     x-api-key: <an api key>
     Authorization: Bearer <access token>
````

File: lib/src/main/java/uk/gov/di/ipv/cri/address/library/service/CredentialIssuerService.java [hunk 1]
```diff
@@ -0,0 +1,84 @@
...
+    public List<CanonicalAddressWithResidency> getAddresses(UUID sessionId)
+            throws CredentialRequestException {
+        var addressSessionItem = dataStore.getItem(sessionId.toString());
+        if (addressSessionItem == null) {
+            throw new CredentialRequestException(ErrorResponse.MISSING_ADDRESS_SESSION_ITEM);
+        }
+        return addressSessionItem.getAddresses();
+    }
+
+    private String getAccessToken(Map<String, String> headers) throws CredentialRequestException {
```
idrop, line 61: After reading https://github.com/alphagov/digital-identity-architecture/blob/main/rfc/0001-crossteam-oauth-profile.md#requests-to-the-protected-resource, its clear that IPV Core MUST send bearer tokens passed in the Authentication header, see [RFC6750](https://datatracker.ietf.org/doc/html/rfc6750). So this check may need some rework, sorry!

infrastructure/lambda/api.yaml
```diff
@@ -142,12 +142,8 @@ paths:
         passthroughBehavior: "when_no_match"
         contentHandling: "CONVERT_TO_TEXT"
         type: "aws_proxy"
+
   /token:
-    parameters:
```
idrop, line 146: could we address this with KBV-300 instead?

infrastructure/lambda/api.yaml
```diff
@@ -186,6 +182,54 @@ paths:
...
+        description: 'A valid access_token (e.g.: authorization: Bearer access-token-value).'
+        schema:
+          type: string
+    post:
+      summary: Post request using a valid access_token
+      description: Accepts form encoded param consisting of the access_token and subject
+
+      responses:
+        '200':
+          description: 201 Ok
```
idrop, line 204: 200 Ok instead?

idrop (approved):
🚀

idrop (approved):

PR #64, opened 2022-04-07
KBV-305 session lambda receive JAR

tal-nagra:
### What changed

The address CRI session lambda function now receives a JWT-secured authorisation request

### Issue tracking


lib/src/main/java/uk/gov/di/ipv/cri/address/library/domain/RawSessionRequest.java
```diff
@@ -0,0 +1,27 @@
+package uk.gov.di.ipv.cri.address.library.domain;
+
+import com.fasterxml.jackson.annotation.JsonAlias;
+
+public class RawSessionRequest {
+    @JsonAlias("client_id")
```
KaraokeStu, line 6: Are we ever going to re-serialize this?
If so, we should use @JsonProperty instead

lib/src/main/java/uk/gov/di/ipv/cri/address/library/domain/RawSessionRequest.java
```diff
@@ -0,0 +1,27 @@
...
+    private String clientId;
+
+    @JsonAlias("request")
```
KaraokeStu, line 9: As above, if we're ever going to re-serialize

KaraokeStu (commented):
Just my comments on re-serializing the object, if we're not going to ever reserialize it, then JsonAlias is fine

lib/src/main/java/uk/gov/di/ipv/cri/address/library/domain/RawSessionRequest.java
tal-nagra: At present, there is no requirement to re-serialise

lib/src/main/java/uk/gov/di/ipv/cri/address/library/domain/RawSessionRequest.java
tal-nagra: At present, there is no requirement to re-serialise

KaraokeStu (approved):

PR #65, opened 2022-04-07
KBV-300: Request the auth token via the private_key_jwt method

idrop:
### What changed
Refactored code to fetch the auth token via the **private_key_jwt** method

### Why did it change
As per the [RFC](https://github.com/alphagov/digital-identity-architecture/blob/main/rfc/0001-crossteam-oauth-profile.md#requests-to-the-token-endpoint), it has been decided that the clients must authenticate to the authorization server's token endpoint using [the private_key_jwt method](https://openid.net/specs/openid-connect-core-1_0.html#ClientAuthentication)

### Issue tracking

- [KBV-300](https://govukverify.atlassian.net/browse/KBV-300)
- [RFC](https://github.com/alphagov/digital-identity-architecture/blob/main/rfc/0001-crossteam-oauth-profile.md#client-profile)


lib/src/main/java/uk/gov/di/ipv/cri/address/library/service/AccessTokenService.java
```diff
@@ -27,18 +41,30 @@
...
-
-    public AccessTokenService(DataStore<AddressSessionItem> dataStore, Long bearerAccessTokenTtl) {
+    private final ConfigurationService configurationService;
+    private final JWTVerifier jwtVerifier;
+
+    public AccessTokenService(
+            DataStore<AddressSessionItem> dataStore,
+            Long bearerAccessTokenTtl,
+            ConfigurationService configurationService,
+            JWTVerifier jwtVerifier) {
```
idrop, line 55: added JWTVerifier as its now used in the AddressSessionService and AddressTokenService

lib/src/main/java/uk/gov/di/ipv/cri/address/library/service/AccessTokenService.java
```diff
@@ -88,41 +90,115 @@ public void writeToken(
...
-                listHelper.getValueOrThrow(
-                        queryParameters.getOrDefault(CODE, Collections.emptyList()));
-        var redirectUri =
-                listHelper.getValueOrThrow(
-                        queryParameters.getOrDefault(REDIRECT_URI, Collections.emptyList()));
-        var grantType =
-                listHelper.getValueOrThrow(
-                        queryParameters.getOrDefault(GRANT_TYPE, Collections.emptyList()));
+            ClientAuthentication clientAuthentication = tokenRequest.getClientAuthentication();
+            PrivateKeyJWT privateKeyJWT = (PrivateKeyJWT) clientAuthentication;
```
idrop, line 119: @amanm-alphagov this is the important bit of the PR - where we get a PrivateKeyJWT, then verify its payload and signing

lib/src/main/java/uk/gov/di/ipv/cri/address/library/service/JWTVerifier.java
```diff
@@ -0,0 +1,98 @@
...
+import java.security.cert.CertificateFactory;
+import java.security.interfaces.ECPublicKey;
+import java.security.interfaces.RSAPublicKey;
+import java.text.ParseException;
+import java.util.Arrays;
+import java.util.Base64;
+import java.util.HashSet;
+import java.util.Map;
+
+public class JWTVerifier {
```
idrop, line 27: @amanm-alphagov sorry but we'll need tests for this new class - you can lift the tests from AddressSessionServiceTest that build up JWTs

lib/src/main/java/uk/gov/di/ipv/cri/address/library/service/AccessTokenService.java
amanm-alphagov: added unit tests for JWTVerifier.class

lib/src/main/java/uk/gov/di/ipv/cri/address/library/service/JWTVerifier.java
amanm-alphagov: added unit tests for JWTVerifier class

lambdas/accesstoken/src/test/java/uk/gov/di/ipv/cri/address/api/handler/AccessTokenHandlerTest.java
```diff
@@ -55,18 +55,25 @@ void setUp() {
...
+        String authCodeValue = "12345";
+        String grantType = "authorization_code";
         String tokenRequestBody =
-                "code=12345&redirect_uri=http://test.com&grant_type=authorization_code&client_id=test_client_id";
+                String.format(
+                        "code=%s"
+                                + "&client_assertion=%s"
+                                + "&client_assertion_type=urn:ietf:params:oauth:client-assertion-type:jwt-bearer"
+                                + "&client_id=urn:uuid:ipv-core"
+                                + "&grant_type=%s",
```
jkunle, line 66: This is really weird as you are using the `String.format` you shouldn't have to use `+`, I think you can replace the `+` with a commas for the same effect

lib/src/main/java/uk/gov/di/ipv/cri/address/library/service/AccessTokenService.java
```diff
@@ -1,83 +1,88 @@
...
-
-    public AccessTokenService(DataStore<AddressSessionItem> dataStore, Long bearerAccessTokenTtl) {
+    private final ConfigurationService configurationService;
+    private final JWTVerifier jwtVerifier;
+    private ListUtil listHelper = new ListUtil();
+
+    public AccessTokenService(
+            DataStore<AddressSessionItem> dataStore,
+            Long bearerAccessTokenTtl,
+            ConfigurationService configurationService,
```
jkunle, line 51: I think `bearerAccessTokenTtl` comes from `configurationService.getBearerAccessTokenTtl` at some stage this class only needed  `bearerAccessTokenTtl` hence there was no reason to pass in the entire configurationService. I think we can remove  `bearerAccessTokenTtl`

jkunle:
Please have look at the bug and code smell on offered by SonarCloud

lib/src/main/java/uk/gov/di/ipv/cri/address/library/service/AccessTokenService.java
```diff
@@ -88,47 +93,119 @@ public void writeToken(
...
+                    "Authorisation code does not match with authorization Code for Address Session Item");
         }
-        if (!URI.create(redirectUri).equals(addressSessionItem.getRedirectUri())) {
-            throw new AccessTokenRequestException(
-                    String.format(
-                            "Requested redirectUri: %s does not match existing redirectUri: %s",
-                            redirectUri, addressSessionItem),
-                    OAuth2Error.INVALID_GRANT);
+
+        Issuer jwtIssuer = privateKeyJWT.getJWTAuthenticationClaimsSet().getIssuer();
```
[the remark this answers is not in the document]
techaman-m: Added a new method with more micro grained level checks.

lib/src/main/java/uk/gov/di/ipv/cri/address/library/service/AccessTokenService.java
techaman-m: Removed.

lambdas/accesstoken/src/test/java/uk/gov/di/ipv/cri/address/api/handler/AccessTokenHandlerTest.java
techaman-m: I think you can use either `+` or `,` depending on the need. The `+` is just a string concatenation operator in this case and it is there to split the string. We can put all the strings one line. Using `,` the same thing can be achieved but the way the params need to be passed will have to change.

jkunle (approved):

PR #66, opened 2022-04-11
KBV-305 - Make AddressCriAudience param value env specific

tal-nagra:
### What changed

The value of the AddressCriAudience SSM parameter (used by the Session lambda function) is assigned on a per environment basis.

### Why did it change
To align with the current credential issuer configuration in di-ipv-config

### Issue tracking


MattBidewell (approved):
LGTM matches what we're doing for other environment variables.

PR #72, opened 2022-04-21
KBV-305 Return state and redirect_uri in session API response

idrop:
## Proposed changes

Return `state` and `redirect_uri` in session API response so that the frontend can save these values to send back in the auth code redirect.

This is now necessary as the Authorization JWT request now contains these values in the Authorization JAR, rather than as http request params.

### Issue tracking

### Other considerations


KaraokeStu (approved):
Looks good

PR #73, opened 2022-04-24
KBV-300 fix token request

idrop:
## Proposed changes

* Log API gateway requests including body payloads
* Remove validation of client_id in token request (it isn't sent)
* Remove validation of `nbf` in token request (it isn't sent)
* Set client id as ipv-core for all dev clients

### Issue tracking

- [KBV-300](https://govukverify.atlassian.net/browse/KBV-300)

### Other considerations

### DIFF
--- a/infrastructure/lambda/api.yaml
+++ b/infrastructure/lambda/api.yaml
@@ -46,7 +46,7 @@ paths:
             application/json:
               schema:
                 $ref: "#/components/schemas/Error"
-      x-amazon-apigateway-request-validator: "Validate body"
+      x-amazon-apigateway-request-validator: "Validate body" ## todo this validator doesn't exist
       x-amazon-apigateway-integration:
         httpMethod: "POST"
         uri:
@@ -149,8 +149,32 @@ paths:
         content:
           application/x-www-form-urlencoded:
             schema:
-              $ref: "#/components/schemas/TokenRequest"
-        required: true
+              required:
+                - "grant_type"
+                - "code"
+                - "client_assertion_type"
+                - "client_assertion"
+                - "redirect_uri"
+              properties:
+                grant_type:
+                  type: "string"
+                  pattern: "authorization_code"
+                  example: "authorization_code"
+                code:
+                  type: "string"
+                  minLength: 1
+                client_assertion_type:
+                  type: "string"
+                  pattern: "urn:ietf:params:oauth:client-assertion-type:jwt-bearer"
+                  example: "urn:ietf:params:oauth:client-assertion-type:jwt-bearer"
+                client_assertion:
+                  type: "string"
+                  pattern: "[a-zA-Z0-9_=]+\\.[a-zA-Z0-9_=]+\\.[a-zA-Z0-9_\\-\\+\\/=]+"
+                  example: "eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJzdWIiOiJ0dCIsImlhdCI6MTUxNjIzOTAyMn0.SbcN-ywpLObhMbbaMCtW1Un8LYhQzHsEth9LvTk4oQQ"
+                redirect_uri:
+                  type: "string"
+                  format: "uri"
+                  example: "https://di-ipv-core-stub.london.cloudapps.digital/callback"
       responses:
         "201":
           description: "201 response"
@@ -170,7 +194,6 @@ paths:
             application/json:
               schema:
                 $ref: "#/components/schemas/Error"
-
       x-amazon-apigateway-request-validator: "Validate both"
       x-amazon-apigateway-integration:
         httpMethod: "POST"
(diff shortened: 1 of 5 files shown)


infrastructure/lambda/api.yaml
```diff
@@ -46,7 +46,7 @@ paths:
             application/json:
               schema:
                 $ref: "#/components/schemas/Error"
-      x-amazon-apigateway-request-validator: "Validate body"
+      x-amazon-apigateway-request-validator: "Validate body" ## todo this validator doesn't exist
```
idrop, line 49: if we set this to `Validate both` then the PUT address call fails - @KaraokeStu could you take a look please?

infrastructure/lambda/api.yaml
```diff
@@ -149,8 +149,32 @@ paths:
...
+                  pattern: "urn:ietf:params:oauth:client-assertion-type:jwt-bearer"
+                  example: "urn:ietf:params:oauth:client-assertion-type:jwt-bearer"
+                client_assertion:
+                  type: "string"
+                  pattern: "[a-zA-Z0-9_=]+\\.[a-zA-Z0-9_=]+\\.[a-zA-Z0-9_\\-\\+\\/=]+"
+                  example: "eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJzdWIiOiJ0dCIsImlhdCI6MTUxNjIzOTAyMn0.SbcN-ywpLObhMbbaMCtW1Un8LYhQzHsEth9LvTk4oQQ"
+                redirect_uri:
+                  type: "string"
+                  format: "uri"
+                  example: "https://di-ipv-core-stub.london.cloudapps.digital/callback"
```
idrop, line 177: inlining the TokenRequest allows the /token call to pass the API gateway - but I think the validation is not working correctly - @jkunle could you take a look at this in a separate ticket please?

infrastructure/lambda/api.yaml
```diff
@@ -195,22 +218,15 @@ paths:
         description: 'A valid access_token (e.g.: authorization: Bearer access-token-value).'
         schema:
           type: string
-    post:
-      summary: Post request using a valid access_token
-      description: Accepts form encoded param consisting of the access_token and subject
-
+    get:
+      summary: GET request using a valid access token
```
idrop, line 222: IPV Core Stub uses the `UserInfoRequest` to build a request to /credential/issue. If we use this class, then we need to use a GET in order to set the token in the `Authorization:Bearer xxxx` header. If we use a POST from this class, the token gets set in the req body

infrastructure/lambda/api.yaml
```diff
@@ -195,22 +218,15 @@ paths:
...
           content:
             application/jwt:
               schema:
-                type: "object"
-                properties:
-                  type: string
-                  format: application/jwt
-                  pattern: ^([a-zA-Z0-9_=]+)\.([a-zA-Z0-9_=]+)\.([a-zA-Z0-9_\-\+\/=]+)$
-                  example: "eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJzdWIiOiIxMjM0NTY3ODkwIiwibmFtZSI6IkpvaG4gRG9lIiwiaWF0IjoxNTE2MjM5MDIyfQ.SflKxwRJSMeKKF2QT4fwpMeJf36POk6yJV_adQssw5c"
+                type: string
```
idrop, line 229: IntelliJ wasn't happy with this block - I think we can just set the content type as application/jwt and body as string

infrastructure/lambda/template.yaml
```diff
@@ -137,7 +137,13 @@ Resources:
         - LoggingLevel: INFO
           ResourcePath: '/*'
           HttpMethod: '*'
+          # Disable data trace in production to avoid logging customer sensitive information
+          DataTraceEnabled: true
+          MetricsEnabled: true
+          ThrottlingRateLimit: 5
+          ThrottlingBurstLimit: 10
       AccessLogSetting:
+        DestinationArn: !Sub 'arn:${AWS::Partition}:logs:${AWS::Region}:${AWS::AccountId}:log-group:${ApiAccessLogGroup}'
```
idrop, line 146: sets data trace enabled logging - we may want to remove this in prod, but the user data in the shared attribute set will be encrypted in any case

infrastructure/lambda/template.yaml
```diff
@@ -155,6 +161,32 @@ Resources:
...
+        Version: "2012-10-17"
+        Statement:
+          - Effect: Allow
+            Principal:
+              Service:
+                - "apigateway.amazonaws.com"
+            Action: "sts:AssumeRole"
+      ManagedPolicyArns:
+        - !Sub "arn:${AWS::Partition}:iam::aws:policy/service-role/AmazonAPIGatewayPushToCloudWatchLogs"
+
```
idrop, line 189: ☝️ supports API gateway logging

lib/src/main/java/uk/gov/di/ipv/cri/address/library/service/AccessTokenService.java
```diff
@@ -104,7 +104,6 @@ public TokenRequest createTokenRequest(String requestBody)
                     .containsAll(
                             Set.of(
                                     CODE,
-                                    CLIENT_ID,
```
idrop, line 107: we don't send a client id in the token req

lib/src/main/java/uk/gov/di/ipv/cri/address/library/service/JWTVerifier.java
```diff
@@ -74,7 +74,7 @@ private void verifyJWTClaimsSet(
                                 .issuer(clientAuthenticationConfig.get("issuer"))
                                 .audience(clientAuthenticationConfig.get("audience"))
                                 .build(),
-                        new HashSet<>(Arrays.asList("exp", "nbf", "sub")));
```
idrop, line 77: we don't send nbf in the token req

infrastructure/lambda/api.yaml
jkunle: If the            ```     type: "object"
                properties:
```
is removed it would be fine, i think this got through during rebase and stuff

idrop: @jkunle If I remove the `type: "object` then IntelliJ seems unhappy with the `properties` block

infrastructure/lambda/template.yaml
jkunle: nice 👍🏾

lib/src/main/java/uk/gov/di/ipv/cri/address/library/service/JWTVerifier.java
tal-nagra: It would be best to add an overload to the public verifyJWT with a requiredClaims HashSet parameter.
Then nbf or anything else can be passed in on a per invocation basis

idrop: We'll handle `nbf` as a separate task

idrop: we should check the nbf for the Auth JAR

PR #76, opened 2022-04-25
Fix canonical address model to match RFC

KaraokeStu:
Canonical address was incorrect - updated to match the RFC


jkunle:
We also need to ensure that the properties of the Canonical address end up in the VC, so we should uncomment the Test in `VerifiableCredentialServiceTest`

jkunle:
We also need to ensure that the properties of the Canonical address end up in the VC, so we should uncomment the Test in `VerifiableCredentialServiceTest`

jkunle (approved):

PR #79, opened 2022-04-26
Make nbf check configurable

techaman-m:
## Proposed changes

Added claims set by creating an enum `RequiredClaims` and changed method signature `verifyJWT` to accept required claims as a list of strings.

### Why did it change

- [KBV-347](https://govukverify.atlassian.net/browse/KBV-347)

### DIFF
--- a/lib/src/main/java/uk/gov/di/ipv/cri/address/library/service/AccessTokenService.java
+++ b/lib/src/main/java/uk/gov/di/ipv/cri/address/library/service/AccessTokenService.java
@@ -148,7 +148,7 @@ public class AccessTokenService {
             jwtVerifier.verifyJWT(
                     clientAuthenticationConfig,
                     signedJWT,
-                    Arrays.asList(RequiredClaims.EXP.name, RequiredClaims.SUB.name));
+                    Arrays.asList(RequiredClaims.EXP.value, RequiredClaims.SUB.value));
             return tokenRequest;
         } catch (SessionValidationException
                 | ClientConfigurationException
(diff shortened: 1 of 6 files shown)


lib/src/main/java/uk/gov/di/ipv/cri/address/library/service/AccessTokenService.java
```diff
@@ -143,7 +145,10 @@ public TokenRequest validateTokenRequest(
                             tokenRequest.getClientAuthentication().getClientID().getValue());
             SignedJWT signedJWT = privateKeyJWT.getClientAssertion();
 
-            jwtVerifier.verifyJWT(clientAuthenticationConfig, signedJWT);
+            jwtVerifier.verifyJWT(
+                    clientAuthenticationConfig,
+                    signedJWT,
+                    Arrays.asList(RequiredClaims.EXP.value, RequiredClaims.SUB.value));
```
jkunle, line 151: Why use `Array.asList` why not `List.of` same for all the occurences of you've used `Array.asList`

jkunle: Please can you remove ` var configurationServiceLocal = new ConfigurationService();` on line 63 and move `this.configurationService = new ConfigurationService();` to position 63

jkunle: I think you should remove    49 and 58 and also `new ListUtil()` shouldn't be in the constructor it is only used once and should be declared in the method where it is used

techaman-m: Addressed all comments

KaraokeStu (approved):
Looks good

PR #81, opened 2022-04-27
KBV-308 - decrypt JAR and verify EC signature

tal-nagra:
### What changed

The session lambda function now decrypts the incoming JAR 
Support has been added for JWT signature verification using the EC signing algorithm

### Issue tracking
- [KBV-308](https://govukverify.atlassian.net/browse/KBV-308)

### DIFF
--- a/lambdas/session/src/main/java/uk/gov/di/ipv/cri/address/api/handler/SessionHandler.java
+++ b/lambdas/session/src/main/java/uk/gov/di/ipv/cri/address/api/handler/SessionHandler.java
@@ -21,7 +21,6 @@ import java.util.Map;
 import java.util.UUID;
 
 import static org.apache.logging.log4j.Level.ERROR;
-import static org.apache.logging.log4j.Level.INFO;
 
 public class SessionHandler
         implements RequestHandler<APIGatewayProxyRequestEvent, APIGatewayProxyResponseEvent> {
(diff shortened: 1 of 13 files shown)


KaraokeStu (approved):
All looks good to me - it may even fix the issue I've been having, but even if it doesn't, I can use the objectmapper to sort the fix quickly

idrop (dismissed):

lambdas/session/src/main/java/uk/gov/di/ipv/cri/address/api/handler/SessionHandler.java
```diff
@@ -70,7 +70,7 @@ public APIGatewayProxyResponseEvent handleRequest(
 
         } catch (SessionValidationException e) {
 
-            eventProbe.log(INFO, e).counterMetric(EVENT_SESSION_CREATED, 0d);
+            eventProbe.log(ERROR, e).counterMetric(EVENT_SESSION_CREATED, 0d);
```
idrop, line 73: I'd say this is still INFO and not an ERROR if we had an issue with the request

tal-nagra: I've never logged an exception with anything other than log level ERROR tbf. I'm not sure if the INFO log level provides the same level of verbosity as the ERROR level

KaraokeStu (approved):
This all looks good now, but will still need my fixes applying to it (should be simple to merge)

PR #86, opened 2022-04-29
KBV-360 Add IPV-Core as a client for staging

idrop:
## Proposed changes

* Add IPV-Core as a client to the Staging Address CRI
* Add additional expected config for integration and staging, accepting that these values may change

Note that clients are distinctly represented by their path in SSM parameter store e.g.
````
/${AWS::StackName}/clients/ipv-core/jwtAuthentication/audience
/${AWS::StackName}/clients/ipv-core-stub/jwtAuthentication/audience
````

This will allow us to accept more than one IPV Core client to support key rotation done by IPV Core, so they could start sending requests with a different client id - which we’d have configured ahead of time.

The `clients/{client-id}/`  portion of the SSM parameter would have to be new. The `1` in the YAML name is a just a way to support this idea that there may be multiple clients from the same issuer.

### Why did it change

Allow IPV Core to integrate with the Address CRI in the stagin environment.

### Issue tracking

### DIFF
--- a/infrastructure/lambda/template.yaml
+++ b/infrastructure/lambda/template.yaml
@@ -23,6 +23,11 @@ Conditions:
   CreateDevResources: !Equals
     - !Ref Environment
     - dev
+  IsStubEnvironment: !Or
+    - !Equals [ !Ref Environment, dev]
+    - !Equals [ !Ref Environment, build ]
+  IsProdLikeEnvironment:
+    !Not [Condition: IsStubEnvironment]
 
 Globals:
   Function:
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: infrastructure/lambda/template.yaml [hunk 1]
```diff
@@ -70,41 +78,46 @@ Mappings:
...
   IPVCoreStubIssuerMapping:
     Environment:
       dev: "ipv-core-stub"
       build: "ipv-core-stub"
-      staging: "https://staging.core.ipv.account.gov.uk"
-      integration: "https://integration.core.ipv.account.gov.uk"
-      production: "https://core.ipv.account.gov.uk"
 
-  IPVCoreStubPublicCertificateToVerifyMapping:
+  IPVCore1IssuerMapping:
```
MattBidewell, line 93: Whats the importance of the '1'?

PR #96, opened 2022-05-05
[KBV-358] Remove secondary indexes on the session table

KaraokeStu:
PR 1 of 4 to recreate secondary indexes on the Session Table to handle correct projections


tal-nagra (approved):
LGTM - step 1, let's remove the `authorizationCode-index`

MattBidewell (approved):

PR #99, opened 2022-05-05
[KBV-358] Re-add secondary index for authorization code on session table

KaraokeStu:
PR 4 of 4 to recreate secondary indexes on the Session Table to handle correct projections


tal-nagra (approved):
Aman's changes now re-instated. Final PR to re-instate authorizationCode-index on SessionTable

PR #102, opened 2022-05-05
KBV-360 transcode KMS DER sigs to concat format

idrop:
## Proposed changes

The EC signatures produced by AWS KMS are in DER format. The Nimbus lib we're using to verify the signatures needs them in concat format.

This is a [copy of the code passport-back](https://github.com/alphagov/di-ipv-cri-uk-passport-back/pull/105/commits/8573927bad3deb68f57bece09e1fc32753222d8e) uses to transcode a DER format signature to concat format.

👉 Also.. fix the redirect uri to a form that IPV Core expects.

### Issue tracking

### DIFF
--- a/infrastructure/lambda/template.yaml
+++ b/infrastructure/lambda/template.yaml
@@ -126,9 +126,9 @@ Mappings:
 
   IPVCore1RedirectURIMapping:
     Environment:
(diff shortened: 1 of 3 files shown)


jkunle (approved):
LGTM

PR #105, opened 2022-05-06
KBV-360 Change the credential/issue method to POST

idrop:
## Proposed changes
Change the credential/issue method to POST, to bring it inline with the method IPV Core is using

### DIFF
--- a/infrastructure/lambda/api.yaml
+++ b/infrastructure/lambda/api.yaml
@@ -218,7 +218,7 @@ paths:
         description: 'A valid access_token (e.g.: authorization: Bearer access-token-value).'
         schema:
           type: string
-    get:
+    post:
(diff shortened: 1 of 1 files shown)


infrastructure/lambda/api.yaml
```diff
@@ -218,7 +218,7 @@ paths:
         description: 'A valid access_token (e.g.: authorization: Bearer access-token-value).'
         schema:
           type: string
-    get:
+    post:
       summary: GET request using a valid access token
```
Danomanic, line 222:
```suggestion
      summary: POST request using a valid access token
```

Danomanic (approved):
Looks good apart from small comment - can be addressed at a late point.

PR #108, opened 2022-05-09
KBV-360 Inline the IssueCredentialFunctionRole

idrop:
## Proposed changes

* Remove the IssueCredentialFunctionRole and map the policy on the IssueCredentialFunction directly
* tidy up KMS sign and SSM policies

### Issue tracking

### DIFF
--- a/infrastructure/lambda/template.yaml
+++ b/infrastructure/lambda/template.yaml
@@ -751,6 +751,43 @@ Resources:
       Principal: apigateway.amazonaws.com
 
 Outputs:
+
+  SessionApiUrl:
+    Description: "URL for the Address API /session resource"
+    Value: !Sub "https://${AddressApi}.execute-api.${AWS::Region}.amazonaws.com/${Environment}/session"
+    Export:
+      Name: !Sub ${AWS::StackName}-SessionApiUrl
+
+  PostcodeLookupApiUrl:
+    Description: "URL for the Address API /postcode-lookup resource"
+    Value: !Sub "https://${AddressApi}.execute-api.${AWS::Region}.amazonaws.com/${Environment}/postcode-lookup"
+    Export:
+      Name: !Sub ${AWS::StackName}-PostcodeLookupApiUrl
+
+  AddressApiUrl:
+    Description: "URL for the Address API /address resource"
+    Value: !Sub "https://${AddressApi}.execute-api.${AWS::Region}.amazonaws.com/${Environment}/address"
+    Export:
+      Name: !Sub ${AWS::StackName}-AddressApiUrl
+
+  TokenApiUrl:
+    Description: "URL for the Address API /token resource"
+    Value: !Sub "https://${AddressApi}.execute-api.${AWS::Region}.amazonaws.com/${Environment}/token"
+    Export:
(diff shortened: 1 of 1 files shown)


infrastructure/lambda/template.yaml
```diff
@@ -346,66 +346,6 @@ Resources:
               Resource:
                 - !Sub "arn:aws:ssm:${AWS::Region}:${AWS::AccountId}:parameter/${AWS::StackName}/clients/*"
 
-  IssueCredentialFunctionRole:
```
idrop, line 349: This role is now inline in the IssueCredentialFunction, like the other lambdas

infrastructure/lambda/template.yaml
```diff
@@ -460,13 +422,6 @@ Resources:
             Action:
               - 'kms:*'
             Resource: '*'
-          - Sid: "Allow use of the key for digital signing"
-            Effect: Allow
-            Principal:
```
idrop, line 465: This is now set on the IssueCredentialFunction

infrastructure/lambda/template.yaml
```diff
@@ -737,13 +692,6 @@ Resources:
       Value: !FindInMap [OrdnanceSurveyAPIURLMapping, Environment, !Ref 'Environment']
       Description: Ordnance Survey Postcode Lookup API URL
 
-  IssueCredentialFunctionPermission:
-    Type: AWS::Lambda::Permission
-    Properties:
-      Action: lambda:InvokeFunction
-      FunctionName: !GetAtt IssueCredentialFunction.Arn
-      Principal: apigateway.amazonaws.com
```
idrop, line 745: moved the permission to where the other permissions are in the template

infrastructure/lambda/template.yaml
```diff
@@ -415,7 +355,29 @@ Resources:
...
+            TableName: !Ref AddressTable
+        - Statement:
+            Effect: Allow
+            Action:
+              - "kms:Sign"
+            Resource: !Sub "arn:aws:kms:${AWS::Region}:${AWS::AccountId}:key/${AddressCriVcSigningKey}"
+        - SQSSendMessagePolicy:
+            QueueName: !ImportValue AuditEventQueueName
+        - SSMParameterReadPolicy:
+            ParameterName: !Sub "${AWS::StackName}/*"
```
idrop, line 373: What do you think of this? These params aren't secret, and I can't think of a securty reason that a lambda shouldn't have access to read these stack-scoped params. If we adopt this on other lambdas it'll reduce boilerplate yaml

KaraokeStu: Any parameters that specifically need locking down should be able to be overridden by IAM so I think we're good, plus we're limiting to our own stack - so we're good there too

infrastructure/lambda/template.yaml
```diff
@@ -460,13 +422,6 @@ Resources:
             Action:
```
KaraokeStu, line 422: Do we really want to be granting access to ALL of KMS functionality for ALL resources (regardless of region, stack, etc?)

I think this should at least be locked down to the stack or at very least environment

jkunle (approved):
We should investigate further the KMS security

PR #136, opened 2022-06-16
BAU - update to use latest common ConfigurationService

tal-nagra:
### What changed

Updated various methods to not refer to the address specific ConfigurationService methods

### Why did it change

So the address specific ConfigurationService methods can be removed.


KaraokeStu (approved):
LGTM

PR #141, opened 2022-06-20
KBV-566 Add a private api gateway

idrop:
## Proposed changes

Add a private api gateway to be used by frontend only.
The public API will be used by IPV Core backchannel OAuth flows calls.

Expose all APIs on both APIGWs for now.

### Issue tracking

- [KBV-566](https://govukverify.atlassian.net/browse/KBV-566)

### DIFF
--- a/infrastructure/lambda/template.yaml
+++ b/infrastructure/lambda/template.yaml
@@ -646,16 +646,16 @@ Resources:
   LinkUsagePlanApiKey1:
     Type: AWS::ApiGateway::UsagePlanKey
     Properties:
-      KeyId: !Ref ApiKey1
+      KeyId: !ImportValue core-infrastructure-ApiKey1
       KeyType: API_KEY
-      UsagePlanId: !Ref ApiUsagePlan
+      UsagePlanId: !Ref PublicAddressApiUsagePlan
 
   LinkUsagePlanApiKey2:
     Type: AWS::ApiGateway::UsagePlanKey
     Properties:
-      KeyId: !Ref ApiKey2
+      KeyId: !ImportValue core-infrastructure-ApiKey2
       KeyType: API_KEY
-      UsagePlanId: !Ref ApiUsagePlan
+      UsagePlanId: !Ref PublicAddressApiUsagePlan
 
   ParameterSessionTableName:
     Type: AWS::SSM::Parameter
--- a/infrastructure/lambda/template.yaml
+++ b/infrastructure/lambda/template.yaml
@@ -222,7 +222,67 @@ Resources:
       EndpointConfiguration:
         Type: REGIONAL
 
-  ApiAccessLogGroup:
+  PrivateAddressApi:
+    Type: AWS::Serverless::Api
+    Properties:
+      Description: Private Address CRI API
+      MethodSettings:
+        - LoggingLevel: INFO
+          ResourcePath: '/*'
+          HttpMethod: '*'
+          # Disable data trace in production to avoid logging customer sensitive information
+          DataTraceEnabled: true
+          MetricsEnabled: true
+          ThrottlingRateLimit: 5
+          ThrottlingBurstLimit: 10
+      AccessLogSetting:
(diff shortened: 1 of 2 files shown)


infrastructure/lambda/template.yaml
```diff
@@ -63,6 +63,15 @@ Globals:
         POWERTOOLS_LOG_LEVEL: INFO
         SQS_AUDIT_EVENT_PREFIX: IPV_ADDRESS_CRI
 Mappings:
+
+  VPCEndpointIdMapping:
```
devdutoitGDS, line 67: assume these will be done via import eventually?

infrastructure/lambda/template.yaml
```diff
@@ -200,10 +222,76 @@ Resources:
...
+          "httpMethod":"$context.httpMethod",
+          "path":"$context.path",
+          "routeKey":"$context.routeKey",
+          "status":"$context.status",
+          "protocol":"$context.protocol",
+          "responseLatency":"$context.responseLatency",
+          "responseLength":"$context.responseLength"
+          }
+      TracingEnabled: true
+      Name: !Sub "address-cri-${AWS::StackName}"
```
devdutoitGDS, line 254: would add private in the name

infrastructure/lambda/template.yaml
```diff
@@ -200,10 +222,76 @@ Resources:
...
+    Type: AWS::Logs::LogGroup
+    Properties:
+      LogGroupName: !Sub /aws/apigateway/${AWS::StackName}-${PublicAddressApi}-public-AccessLogs
+      RetentionInDays: 365
+
+  PrivateAddressApiAccessLogGroup:
     Type: AWS::Logs::LogGroup
     Properties:
-      LogGroupName: !Sub /aws/apigateway/AccessLog-${AddressApi}/${Environment}
+      LogGroupName: !Sub /aws/apigateway/${AWS::StackName}-${PrivateAddressApi}-private-AccessLogs
```
devdutoitGDS, line 294: if you change the `Name` of `PrivateAddressApi` then you can drop the private in the loggroupname here (same for public)

infrastructure/lambda/template.yaml
```diff
@@ -778,50 +869,38 @@ Resources:
...
-  PostcodeLookupApiUrl:
-    Description: "URL for the Address API /postcode-lookup resource"
-    Value: !Sub "https://${AddressApi}.execute-api.${AWS::Region}.amazonaws.com/${Environment}/postcode-lookup"
-    Export:
-      Name: !Sub ${AWS::StackName}-PostcodeLookupApiUrl
-
-  AddressApiUrl:
-    Description: "URL for the Address API /address resource"
-    Value: !Sub "https://${AddressApi}.execute-api.${AWS::Region}.amazonaws.com/${Environment}/address"
+  AddressApiBaseUrl:
```
devdutoitGDS, line 872: would rename this with `Public` in the name, or at least add a second one with `Public` for now, and then moves things (dns/FE) over to the new one.

devdutoitGDS: Just saw that is exactly what you did :)

devdutoitGDS (dismissed):

infrastructure/lambda/template.yaml
idrop: 👍

infrastructure/lambda/template.yaml
idrop: yeah @mrdoodles  is exporting them in a VPC stack, but they're not available in all envs yet

infrastructure/lambda/template.yaml
idrop: @devdutoitGDS If I change this to `!Sub /aws/apigateway/${AWS::StackName}-${PrivateAddressApi}-AccessLogs`then my loggroup in dev looks like `/aws/apigateway/cri-address-philm-315b7dan46-AccessLogs` which doesn't include a public or private part. What should I use in this `!Sub` expression?

devdutoitGDS: hmm, is this after adding `private` to `Name:` in `PrivateAddressApi` ?

devdutoitGDS (approved):

PR #145, opened 2022-06-23
KBV-593: Separate public private endpoints

jkunle:
## Proposed changes
This story is to ensure that only public endpoints are in the `api.yaml` to be renamed `public-api.yaml` and only private endpoints should be specified in `private-api.yaml` 

### What changed

- Rename `api.yaml` to `public-api.yaml`
- Remove private endpoints in `private-api.yaml` 
- Remove public endpoints in `public-api.yaml`

### Why did it change

To ensure clarity

- [KBV-593](https://govukverify.atlassian.net/browse/KBV-593)


infrastructure/lambda/template.yaml
```diff
@@ -632,21 +632,6 @@ Resources:
...
-      ApiStages:
-        - ApiId: !Ref PrivateAddressApi
-          Stage: !Ref Environment
-      Quota:
-        Limit: 500000
-        Period: DAY
-      Throttle:
-        BurstLimit: 100 # requests the API can handle concurrently
-        RateLimit: 50 # allowed requests per second
-
```
idrop, line 649: I think we should retain this usage plan as it protects our API from being spammed by calls to the frontend

idrop (approved):

PR #200, opened 2022-10-07
Add github actions for tests and linting

MattBidewell:
### What changed

Added github actions to lint and run the tests of the nodelambda functions.

It would be cool if we got around to use precommit in our github actions to run all our checks.

### Why did it change

Runs linter and tests when we create a PR.

### Issue tracking

- [OJ-958](https://govukverify.atlassian.net/browse/OJ-958)

### DIFF
--- a/.github/workflows/pre-merge-checks.yml
+++ b/.github/workflows/pre-merge-checks.yml
@@ -8,6 +8,7 @@ on:
       - synchronize
 
 jobs:
+
   style-checks:
     runs-on: ubuntu-latest
     steps:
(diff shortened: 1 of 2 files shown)


.github/workflows/pre-merge-checks.yml
```diff
@@ -22,6 +23,23 @@ jobs:
       - name: Run Spotless
         run: ./gradlew --no-daemon spotlessCheck
 
+  lint:
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        node-function:
+          - get-addresses
```
MattBidewell, line 31: Strategy: Matrix: allows you to set variables that are then reused in the steps. 
The steps will run for same number of strategy that exist. In this use we would add a new node-function value whenever we add a new function.

idrop (approved):

jkunle (approved):

PR #206, opened 2022-10-12
OJ-992-use-param-store

MattBidewell:
### What changed

Use paramstore for the table name rather than a config file.

### Why did it change

We should use Param store when using configuration variables like table names. This will allow us to change it via infrastructure if needed.

### Issue tracking

- [OJ-992](https://govukverify.atlassian.net/browse/OJ-992)

- [X] No environment variables or secrets were added or changed

### DIFF
--- a/infrastructure/lambda/template.yaml
+++ b/infrastructure/lambda/template.yaml
@@ -520,18 +520,24 @@ Resources:
       Environment:
         Variables:
           POWERTOOLS_SERVICE_NAME: !Sub "${CriIdentifier}-get-addresses"
-          PERSON_IDENTITY_TABLE_NAME: !Sub "{{resolve:ssm:/${CommonStackName}/PersonIdentityTableName}}"
       Policies:
         - AWSLambdaBasicExecutionRole
         - AWSXrayWriteOnlyAccess
         - DynamoDBReadPolicy:
             TableName: !Sub "{{resolve:ssm:/${CommonStackName}/PersonIdentityTableName}}"
+        - Statement:
+          - Sid: ReadParametersPolicy
+            Effect: Allow
+            Action:
+              - ssm:GetParameter
+            Resource:
+              - !Sub "arn:aws:ssm:${AWS::Region}:${AWS::AccountId}:parameter/${AWS::StackName}/AddressLookupTableName"
     Metadata: # Manage esbuild properties
       BuildMethod: esbuild
       BuildProperties:
         Minify: true
         Target: "es2020"
-        # Sourcemap: true # Enabling source maps will create the required NODE_OPTIONS environment variables on your lambda function during sam build
+        Sourcemap: true # Enabling source maps will create the required NODE_OPTIONS environment variables on your lambda function during sam build
         EntryPoints:
         - src/app.ts
 
@@ -632,6 +638,14 @@ Resources:
       Type: String
       Value: !FindInMap [OrdnanceSurveyAPIURLMapping, Environment, !Ref 'Environment']
       Description: Ordnance Survey Postcode Lookup API URL
+
+  AddressLookupTableNameParameter:
+    Type: AWS::SSM::Parameter
+    Properties:
+      Name: !Sub "/${AWS::StackName}/AddressLookupTableName"
+      Type: String
+      Value: !Sub "{{resolve:ssm:/${CommonStackName}/PersonIdentityTableName}}"
+      Description: Table name to retrieve addresses from store
 
   PostcodeLookupFunctionPermission:
     Type: AWS::Lambda::Permission
--- a/lambdas/get-addresses/src/app.ts
+++ b/lambdas/get-addresses/src/app.ts
@@ -1,11 +1,17 @@
 import { APIGatewayProxyEvent, APIGatewayProxyResult } from "aws-lambda";
 import { AddressService } from "./services/address-service";
-import config from "./config";
 import { DynamoDbClient } from "./lib/dynamo-db-client";
+import { SsmClient } from "./lib/param-store-client";
+import { ConfigService } from "./services/config-service";
+
+const configService = new ConfigService(SsmClient);
+const initPromise = configService.init();
 
 export const lambdaHandler = async (event: APIGatewayProxyEvent): Promise<APIGatewayProxyResult> => {
     let response: APIGatewayProxyResult;
     try {
+        await initPromise;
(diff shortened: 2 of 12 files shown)


infrastructure/lambda/template.yaml
```diff
@@ -566,6 +574,13 @@ Resources:
         - AWSXrayWriteOnlyAccess
         - DynamoDBReadPolicy:
             TableName: !Sub "{{resolve:ssm:/${CommonStackName}/PersonIdentityTableName}}"
+        - Statement:
```
MattBidewell, line 577: We could see it SAM Connectors could be used to simplify permissions - 

https://docs.aws.amazon.com/serverless-application-model/latest/developerguide/managing-permissions-connectors.html

lambdas/get-addresses/src/app.ts
```diff
@@ -1,11 +1,17 @@
 import { APIGatewayProxyEvent, APIGatewayProxyResult } from "aws-lambda";
 import { AddressService } from "./services/address-service";
-import config from "./config";
 import { DynamoDbClient } from "./lib/dynamo-db-client";
+import { SsmClient } from "./lib/param-store-client";
+import { ConfigService } from "./services/config-service";
+
+const configService = new ConfigService(SsmClient);
+const initPromise = configService.init();
```
MattBidewell, line 8: initialise the config service promise outside the lambdaHandler. This will allow Lambda to cache the ssm results.

However we would need to resolve it inside the handler. (very painful to test)

lambdas/get-addresses/src/services/config-service.ts
```diff
@@ -0,0 +1,37 @@
...
+    private parameterPrefix = process.env.AWS_STACK_NAME || "";
+    readonly config: { addressLookupStorageTableName: string | undefined } = {
+        addressLookupStorageTableName: undefined,
+    };
+
+    public async init(): Promise<void[]> {
+        // how do handle the parameter names
+        const getParameterAddressQueryTable = this.getParameter(
+            `${this.parameterPrefix}/GetAddressQueryTable`,
+            "addressLookupStorageTableName",
```
MattBidewell, line 15: This can be simplified if we use the same name for the parameter store and what gets stored as a key in the `config` variable. Opinions?

lambdas/get-addresses/src/services/config-service.ts
```diff
@@ -0,0 +1,37 @@
...
+        );
+
+        return Promise.all([getParameterAddressQueryTable]);
```
MattBidewell, line 18: Future-proofing this, if we ever add more variables we can promise.all them to keep `init` speed high.

lambdas/get-addresses/src/services/config-service.ts
```diff
@@ -0,0 +1,41 @@
...
+            Name: paramName,
+        });
+
+        const addressLookupStorageTableName = await this.ssmClient.send(command);
+        const value = addressLookupStorageTableName?.Parameter?.Value;
+        const ObjKey = key as keyof Config;
+
+        if (!value) {
+            // eslint-disable-next-line no-console
+            console.warn(`Undefined parameter - ${key}`);
```
MattBidewell, line 36: If the parameter is undefined. Should we
- Throw an error and process exit 1.
- Log that its missing and attempt to continue.

infrastructure/lambda/template.yaml
idrop: Oh cool! Like:
````
  MyConnector:
    Type: AWS::Serverless::Connector
    Properties:
      Source:
        Id: MyFunction
      Destination:
        Id: MyTable
      Permissions:
        - Write
       ````

**Review by idrop — DISMISSED** [review]

**Review by mrdoodles — DISMISSED** [review]

File: lambdas/get-addresses/src/services/config-service.ts [hunk 6]
```diff
@@ -0,0 +1,40 @@
...
+        const command = new GetParameterCommand({
+            Name: paramName,
+        });
+
+        const addressLookupStorageTableName = await this.ssmClient.send(command);
+        const value = addressLookupStorageTableName?.Parameter?.Value;
+        const ObjKey = key as keyof Config;
+
+        if (!value) {
+            throw new Error(`Missing Parameter - ${key}`);
```
MattBidewell, line 35: You cant put `Promise.reject` here as that creates an unhandled promise rejection inside the promise. 

Trowing an error will correctly promise.reject the promise instead.

If we was creating a Promise manually, you would use Promise.reject instead

lambdas/get-addresses/src/lib/param-store-client.ts
```diff
@@ -0,0 +1,3 @@
+import { SSMClient } from "@aws-sdk/client-ssm";
+
+export const SsmClient = new SSMClient({ region: "eu-west-2" });
```
jkunle, line 3: Should we be explicitly setting the region here, can we maybe use AWS_REGION environment variable see: https://docs.aws.amazon.com/sdk-for-javascript/v3/developer-guide/setting-region.html

MattBidewell: we could use an environment variable to set the region, but we can only run out of eu-west-2 so not sure what setting it via an env would give us

lambdas/get-addresses/src/app.ts
```diff
@@ -15,7 +21,7 @@ export const lambdaHandler = async (event: APIGatewayProxyEvent): Promise<APIGat
             return response;
         }
 
-        const addressService = new AddressService(config.addressLookupStorageTableName, DynamoDbClient);
+        const addressService = new AddressService(configService.config.AddressLookupTableName, DynamoDbClient);
```
jkunle, line 24: Just wondering if we could move this outside the lambdaHandler to maybe line 9 for instance. Just lessons from the java lambda world where is it adviced to instantiate outside the handler code.

MattBidewell: we identified this in our run through. We decided to keep it like this for now as the overhead is tiny compared to Java, however we will come back to address it. 

moving the class initialisations out of the handler makes it a bit of a pain to test.

merlinc (approved):
LGTM

jkunle (approved):
This is good enough, i'm still a little nervous about the region thing, for now it is not a problem for disaster recovery purposes it could be an issue.

PR #224, opened 2022-10-24
sum apigw errors across public and private gateways

idrop:
## Proposed changes

For 5xx and 4x errors, use the sum of errors across the public and private API gateways.
Previously we used the sum of the API stage, and this didn't seem to work.

The sum expression `e1` is the only metric with `ReturnData: true` so this is the single metric that makes up the alarm.


merlinc (approved):
LGTM

mrdoodles (approved):

PR #409, opened 2023-01-25
BAU: Describing parameters needed for Authorization endpoint

jkunle:
### What changed

A fuller description of the url parameters needed for the Authorization endpoint

### Why did it change

To add more documentation for the Authorization when looking at the OpenApi file


MattBidewell (approved):

merlinc (approved):
LGTM - this matches what's being used by the [frontend common oauth code](https://github.com/alphagov/di-ipv-cri-common-express/blob/main/src/routes/oauth2/middleware.js#L71-L81).

At some point we should update the [frontend mocks](https://github.com/alphagov/di-ipv-cri-address-front/blob/main/test/mocks/mappings/addresses.json) to assert these values are passed

PR #492, opened 2023-03-10
Rename new lambdas from two to TS

scronk1:
### What changed

Provide more clarity for us and other teams
Better practice

### Why did it change

API calls would error otherwise


surajkumar (approved):
Looks good to me

scronk1:
Change covered elsewhere

PR #879, opened 2024-02-01
Inciden-660: use power tools to cache parameters

jkunle:
## Why

see https://govukverify.atlassian.net/browse/INCIDEN-660

Similar to https://github.com/govuk-one-login/ipv-spot/pull/951

### What changed

Introduce parameter caching using the default 5mins provided by powertools. 

### Why did it change

To resolve ERROR InvalidSignatureException: Signature expired: 20240129T194125Z is now earlier than 20240129T194717Z (20240129T195217Z - 5 min.)

### Issue tracking

- [Inciden-660](https://govukverify.atlassian.net/browse/Inciden-660)


lambdas/get-addresses/src/app.ts
```diff
@@ -1,42 +1,31 @@
...
 import { DynamoDbClient } from "./lib/dynamo-db-client";
-import { SsmClient } from "./lib/param-store-client";
-import { ConfigService } from "./services/config-service";
+import { getParameter } from "@aws-lambda-powertools/parameters/ssm";
+import { Logger } from "@aws-lambda-powertools/logger";
 
-const configService = new ConfigService(SsmClient);
-const initPromise = configService.init();
+const logger = new Logger();
+const PARAMETER_PREFIX = process.env.AWS_STACK_NAME || "";
```
CharlesIC, line 8:
```suggestion
const parameterPrefix = process.env.AWS_STACK_NAME || "";
```
We can just use js/ts case for variables defined in the code.

lambdas/get-addresses/src/app.ts
```diff
@@ -1,42 +1,31 @@
...
+
+const apiResponse = (statusCode: number, body: string): APIGatewayProxyResult => ({
+    statusCode,
+    body,
+});
```
CharlesIC, line 13:
```suggestion
```
This is useful but should be defined separately in a utils file. But for this function we probably don't need it anyway.

lambdas/get-addresses/src/app.ts
```diff
@@ -1,42 +1,31 @@
...
-
-        const sessionId = event.headers["session_id"] as string;
+        const sessionId = event.headers["session_id"];
         if (!sessionId) {
-            response = {
-                statusCode: 400,
-                body: "Missing header: session_id is required",
-            };
-            return response;
+            return apiResponse(400, "Missing header: session_id is required");
```
CharlesIC, line 19:
```suggestion
            return { statusCode: 400, body: "Missing header: session_id is required" };
```

lambdas/get-addresses/src/services/config-service.ts
CharlesIC, line 1: 🎉

lambdas/get-addresses/tests/unit/app.test.ts
```diff
@@ -1,26 +1,19 @@
...
-                    AddressLookupTableName: "Test",
-                },
-            };
-        }),
-    };
-});
-
+jest.mock("@aws-lambda-powertools/parameters/ssm", () => ({
+    getParameter: jest.fn(),
+}));
```
CharlesIC, line 3: This can be moved below the imports - jest automatically hoists mocking calls to the top when running tests.

lambdas/get-addresses/tests/unit/app.test.ts
```diff
@@ -1,26 +1,19 @@
...
+import { getParameter } from "@aws-lambda-powertools/parameters/ssm";
 import { APIGatewayProxyEvent } from "aws-lambda";
 import { lambdaHandler } from "../../src/app";
 import { CanonicalAddress } from "../../src/types/address";
 import { DynamoDbClient } from "../../src/lib/dynamo-db-client";
 
 const mockDynamoDbClient = jest.mocked(DynamoDbClient);
+const mockedGetParameter = getParameter as jest.MockedFunction<typeof getParameter>;
```
CharlesIC, line 11:
```suggestion
const dynamoDbClientMock = jest.mocked(DynamoDbClient);
const getParameterMock = getParameter as jest.MockedFunction<typeof getParameter>;
```
Both should be either `mocked...` or `mock...` for consitency - or better `...Mock`

CharlesIC (commented):
The changes look good - just some style comments

surajkumar (approved):

PR #1034, opened 2024-08-02
OJ-2650: Run Address VC Pact test via Github Action

jkunle:
## Proposed changes
see: https://govukverify.atlassian.net/browse/OJ-2650

Added relevant secrets to Github, use the Pact Broker


.github/workflows/run-pact-tests.yml
```diff
@@ -0,0 +1,69 @@
...
+      - name: Run pact tests
+        env:
+          PACT_BROKER_HOST: ${{ secrets.pact-broker-host }}
+          PACT_BROKER_USERNAME: ${{ secrets.pact-broker-username }}
+          PACT_BROKER_PASSWORD: ${{ secrets.pact-broker-password }}
+        run: ./gradlew :lambdas:issuecredential:pactTests -x spotlessApply -x spotlessCheck
+
+      - name: Upload test report
+        uses: actions/upload-artifact@v4
+        if: ${{ always() }}
```
Caitlin-cooling, line 62: Do we need to do this always or can we just do it on failure?

PR #1049, opened 2024-08-28
OJ-2745: Add API response metrics and logging

surajkumar:
### What changed
`PostcodeLookupService` added a metric `lookup_postcode_duration` to measure the latency between third-party API calls.

### Why did it change
This change was added so that we have visibility of how long our API calls to third-parties takes. Primarily, this is to aid bottlenecks in our system.

**Screenshot**
image

### Issue tracking

- [OJ-2745](https://govukverify.atlassian.net/browse/OJ-2745)

[OJ-2745]: https://govukverify.atlassian.net/browse/OJ-2745?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/lambdas/postcode-lookup/src/main/java/uk/gov/di/ipv/cri/address/api/service/PostcodeLookupService.java
+++ b/lambdas/postcode-lookup/src/main/java/uk/gov/di/ipv/cri/address/api/service/PostcodeLookupService.java
@@ -6,6 +6,7 @@ import org.apache.logging.log4j.Logger;
 import software.amazon.awssdk.http.HttpStatusCode;
 import software.amazon.awssdk.http.SdkHttpFullRequest;
 import software.amazon.awssdk.http.SdkHttpMethod;
+import software.amazon.cloudwatchlogs.emf.model.Unit;
 import software.amazon.lambda.powertools.tracing.Tracing;
 import uk.gov.di.ipv.cri.address.api.exceptions.PostcodeLookupBadRequestException;
 import uk.gov.di.ipv.cri.address.api.exceptions.PostcodeLookupProcessingException;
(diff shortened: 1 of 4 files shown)


lambdas/postcode-lookup/src/main/java/uk/gov/di/ipv/cri/address/api/service/PostcodeLookupService.java
```diff
@@ -73,7 +79,18 @@ public List<CanonicalAddress> lookupPostcode(String postcode)
         this.validatePostCode(postcode);
         // Create our http request
         HttpRequest request = createHttpRequest(postcode);
+
+        long startTime = System.nanoTime();
```
jkunle, line 83: I think this is nicer

```
Instant start = Instant.now();
doSomething()
Instant start = Instant.now();
Duration.between(start, end).toMillis()
```

surajkumar: Maybe but there's a fundamental problem with `Instant` which is that it's based on the system clock so it can be affected by factors such as manual adjustments or daylight savings changes. I chose `System.nanoTime()` because it's based on a monotonous clock that is not affected by system time changes, making it more reliable for measuring elapsed time. It is designed to measure elapsed time and is guaranteed to be monotonic, meaning it always moves forward and is not subject to changes in the system clock.

The same issues occurs with `System.currentTimeMillis()` and it was one of the factors that was considered in check-hmrc hence why we used `performance.now()` there.

jkunle (approved):

PR #1079, opened 2024-10-17
OJ-2817: Make country code required

mikebeeby:
### What changed

- Removed setAddressCountryIfMissing logic
- Throw AddressProcessingException and return a 500 if an address has no country code. 
- Added unit tests
- Added happy and unhappy integration tests to handle country code's being provided or not. 
- There is also a temporary integration test for a none GB country code. This should be updated when we can handle none UK postcode journeys

### Why did it change

Previously the AddressHandler would add the country code GB to any addresses without country code. This will now enforce the country code to be set in the request body for each address.

Postcodes selected from the OrdnanceSurvey will still have GB set through the PostCodeLookupHandler.  

### Issue tracking

- [OJ-2817](https://govukverify.atlassian.net/browse/OJ-2817)

[OJ-2817]: https://govukverify.atlassian.net/browse/OJ-2817?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### Evidence

image

Changing hardcoded value in address-front to `LOL`
image

Removing hardcoded country code field from FE for manually inputted address.
image


integration-tests/src/test/resources/features/AddressAPIHappyPath.feature
```diff
@@ -69,3 +69,35 @@ Feature: Address API happy path test
...
+    And a valid authorization code is returned in the response
+
+    # Access token
+    When user sends a POST request to token end point
+    And a valid access token code is returned in the response
+
+    # Credential Issued
+    When user sends a POST request to Credential Issue end point with a valid access token
+    And a valid JWT is returned in the response
+    And 5 events are deleted from the audit events SQS queue
```
Caitlin-cooling, line 103: FYI @MarieseMikely as this is related to the work you're currently doing - might want to rebase after this has been merged if you're changing any of the features

Caitlin-cooling (approved):
Nice stuff

PR #1086, opened 2024-10-21
OJ-2813: address v2 code

jkunle:
## Proposed changes

Implement V2 code changes, that will be used to generate the expected structure for the `/addresses/v2` endpoint
Subsequent PR's will connect this up with `/addresses/v2` endpoint

Once merged, am going to continue on the work [here](https://github.com/govuk-one-login/ipv-cri-address-api/pull/1080)

```
{
  addresses: [
    {
      validFrom?: string;
      validUntil?: string;
      uprn?: string;
      organisationName?: string;
      departmentName?: string;
      subBuildingName?: string;
      buildingName?: string;
      buildingNumber?: string;
      dependentStreetName?: string;
      streetName?: string;
      doubleDependentAddressLocality?: string;
      dependentAddressLocality?: string;
      addressLocality?: string;
      postalCode?: string;
      addressCountry?: string;
    }
  ], // empty array if no addresses
  context?: string;
}
```

### What changed

When receiving a context claim in the JAR when a user starts Address CRI.  A new version of the API is needed that would be released to also return the context. separate from the current which doesn't this will leave side by side for a while

### Why did it change

We receive a context claim in the JAR when a user starts Address CRI. We need a way for the fronted to receive this claim so that it can route them to the correct page. We already make an API call to fetch shared_claims that come from this JAR

- [OJ-2813](https://govukverify.atlassian.net/browse/OJ-2813)

[OJ-2813]: https://govukverify.atlassian.net/browse/OJ-2813?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/lambdas/get-addresses/src/lib/error-handler.ts
+++ b/lambdas/get-addresses/src/lib/error-handler.ts
@@ -1,0 +1,26 @@
+import { Logger } from "@aws-lambda-powertools/logger";
+import { APIGatewayProxyResult } from "aws-lambda";
+
+export const handleError = (logger: Logger, error: unknown, functionName: string): APIGatewayProxyResult => {
+    let statusCode = 500;
+    let body = "Internal Server Error";
+
+    if (error instanceof Error) {
+        const { message, stack } = error;
+        body = (error as unknown as APIGatewayProxyResult).body || message;
+        statusCode = (error as unknown as APIGatewayProxyResult).statusCode || 500;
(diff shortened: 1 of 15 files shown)


surajkumar (approved):

jkunle:
Would look in the hacky error-handler in the next PR

PR #1099, opened 2024-10-28
OJ-2810: Add ttl to Address Table

mikebeeby:
### What changed

Added expiryDate to AddressItem and set as TTL
Use the SessionTtl for the AddressTtl as these values should be the same. 

### Why did it change

We should only keep data for as long as we need it (especially as it’s PII). Without a TTL to remove it after some time means that we’ll be accumulating unnecessary costs and also eventually breach GDPR.

### Issue tracking

- [OJ-2810](https://govukverify.atlassian.net/browse/OJ-2810)

### Evidence

image

[OJ-2810]: https://govukverify.atlassian.net/browse/OJ-2810?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/infrastructure/lambda/template.yaml
+++ b/infrastructure/lambda/template.yaml
@@ -672,6 +672,9 @@ Resources:
       KeySchema:
         - AttributeName: "sessionId"
           KeyType: "HASH"
+      TimeToLiveSpecification:
+        AttributeName: expiryDate
+        Enabled: true
 
   PublicAddressApiUsagePlan:
     Type: AWS::ApiGateway::UsagePlan
--- a/lib/src/main/java/uk/gov/di/ipv/cri/address/library/persistence/item/AddressItem.java
+++ b/lib/src/main/java/uk/gov/di/ipv/cri/address/library/persistence/item/AddressItem.java
@@ -12,6 +12,7 @@ import java.util.UUID;
 @DynamoDbBean
 public class AddressItem {
     private UUID sessionId;
+    private long expiryDate;
     private List<CanonicalAddress> addresses = new ArrayList<>();
 
     @DynamoDbPartitionKey()
@@ -21,6 +22,14 @@ public class AddressItem {
 
     public void setSessionId(UUID sessionId) {
         this.sessionId = sessionId;
+    }
+
+    public long getExpiryDate() {
+        return expiryDate;
+    }
+
+    public void setExpiryDate(long expiryDate) {
+        this.expiryDate = expiryDate;
     }
 
     public List<CanonicalAddress> getAddresses() {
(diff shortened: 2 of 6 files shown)


infrastructure/lambda/template.yaml
```diff
@@ -759,6 +763,14 @@ Resources:
       Value: !Sub "{{resolve:ssm:/${CommonStackName}/PersonIdentityTableName}}"
       Description: Table name to retrieve addresses from store
 
+  AddressTtlParameter:
+    Type: AWS::SSM::Parameter
+    Properties:
+      Name: !Sub "/${AWS::StackName}/AddressTtl"
```
Caitlin-cooling, line 769: I guess this is following existing patterns by using an SSM param, but we have SSM rate limiting issues. Do you think it really needs to be one or could it be a constant?

lib/src/main/java/uk/gov/di/ipv/cri/address/library/persistence/item/AddressItem.java
```diff
@@ -23,6 +25,22 @@ public void setSessionId(UUID sessionId) {
         this.sessionId = sessionId;
     }
 
+    public long getCreatedDate() {
```
Caitlin-cooling, line 28: Out of interest what do we need the createdDate for?

infrastructure/lambda/template.yaml
mikebeeby: Nope, I just followed the existing patterns. I'm happy to move it just to be a constant in the code

lib/src/main/java/uk/gov/di/ipv/cri/address/library/persistence/item/AddressItem.java
mikebeeby: Similar to the above I followed an existing implementation, I presumed it may be best practice incase there are any issues with the TTL. 

You're probably right though, no real need for us to have it here for such a short TTL.

infrastructure/lambda/template.yaml
Caitlin-cooling: Oh actually, the TTL for this table should always be the same as the session and person identity tables because it's all the same data - if a user has a record in the address table but not the session table they're not going to be able to do anything. Shall we just reuse the existing `SessionTtl` SSM param? 

In the mid to long term future we will be migrating a lot of these SSM params to app config ([ADR 188](https://github.com/govuk-one-login/architecture/blob/26245cddb0b81326498e3df4a5a75b2633297c8e/adr/0188-application-configuration-feature-flags.md)) so probably makes sense just to follow the patterns laid out for now

lib/src/main/java/uk/gov/di/ipv/cri/address/library/persistence/item/AddressItem.java
Caitlin-cooling: Ok cool fine to leave it in I think 👍

Caitlin-cooling (approved):
Nice stuff looks good to me

PR #1158, opened 2025-01-09
OJ-2974: Add isUkAddress to VC_ISSUED audit event

mikebeeby:
### What changed

- Added isUkAddress to VC_ISSUED audit event extensions, this is set to true for UK, IoM and channel islands. 
- Added VC_ISSUED json schema validation to integration tests.
- Fix: Added context field to START event schema validation

### Why did it change

So that we can differentiate between UK and international addresses for analytics.

### Issue tracking
- [OJ-2974](https://govukverify.atlassian.net/browse/OJ-2974)

[OJ-2974]: https://govukverify.atlassian.net/browse/OJ-2974?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/integration-tests/src/test/resources/schema/IPV_ADDRESS_CRI_START.json
+++ b/integration-tests/src/test/resources/schema/IPV_ADDRESS_CRI_START.json
@@ -50,6 +50,22 @@
         }
       },
       "required": ["device_information"]
+    },
+    "extensions": {
+      "type": "object",
+      "properties": {
+        "evidence": {
+          "type": "array",
+          "items": {
+            "type": "object",
+            "properties": {
+              "context": {
+                "type": "string"
+              }
+            }
+          }
+        }
+      }
     }
   },
   "required": [
--- a/integration-tests/src/test/resources/schema/IPV_ADDRESS_CRI_VC_ISSUED.json
+++ b/integration-tests/src/test/resources/schema/IPV_ADDRESS_CRI_VC_ISSUED.json
@@ -1,0 +1,85 @@
+{
+  "type": "object",
+  "properties": {
+    "event_name": {
+      "type": "string"
+    },
+    "user": {
+      "type": "object",
+      "properties": {
+        "user_id": {
+          "type": "string"
+        },
+        "session_id": {
+          "type": "string"
+        },
+        "persistent_session_id": {
+          "type": "string"
+        },
+        "govuk_signin_journey_id": {
+          "type": "string"
+        },
+        "ip_address": {
+          "type": "string"
+        }
+      },
+      "required": [
+        "user_id",
+        "session_id",
+        "persistent_session_id",
+        "govuk_signin_journey_id",
+        "ip_address"
+      ]
+    },
+    "timestamp": {
+      "type": "integer"
+    },
+    "event_timestamp_ms": {
+      "type": "integer"
+    },
+    "component_id": {
+      "type": "string"
+    },
+    "restricted": {
+      "type": "object",
+      "properties": {
+        "address": {
+          "type": "array",
+          "items": {
+            "type": "object",
+            "properties": {
+              "addressCountry": {
+                "type": "string"
+              }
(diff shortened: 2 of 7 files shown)


integration-tests/src/test/resources/schema/IPV_ADDRESS_CRI_START.json
```diff
@@ -50,6 +50,22 @@
         }
       },
       "required": ["device_information"]
+    },
+    "extensions": {
```
Caitlin-cooling, line 54: Ah nice

integration-tests/src/test/resources/schema/IPV_ADDRESS_CRI_VC_ISSUED.json
```diff
@@ -0,0 +1,81 @@
+{
+  "type": "object",
+  "properties": {
+    "event_name": {
+      "type": "string"
+    },
+    "user": {
+      "type": "object",
+      "properties": {
```
Caitlin-cooling, line 9: Just looking at the [event catalogu](https://github.com/govuk-one-login/event-catalogue/blob/687c0434db77bcb58a013b572ebbf90373175fbb/schemas/IPV_ADDRESS_CRI_VC_ISSUED.yaml#L29)e and we have `persistent_session_id` in there too. Is there a reason we don't need it here?

lambdas/issuecredential/src/main/java/uk/gov/di/ipv/cri/address/api/service/VerifiableCredentialService.java
```diff
@@ -67,12 +68,27 @@ public Map<String, Object> getAuditEventExtensions(List<CanonicalAddress> addres
...
+        if (addresses == null
+                || addresses.get(0) == null
+                || addresses.get(0).getAddressCountry() == null) {
+            return "";
+        }
+        return addresses.get(0).getAddressCountry();
+    }
+
+    private boolean isUkAddress(final String countryCode) {
+        return Stream.of("GB", "GG", "JE", "IM").anyMatch(countryCode::equalsIgnoreCase);
```
Caitlin-cooling, line 92: Can I just check - is there a way that the frontend would currently send GG, JE or JM? My thinking was no, because we send them to the UK address form which hard codes country code GB

Not suggesting we remove this it just plays into another conversation I'm having

integration-tests/src/test/resources/schema/IPV_ADDRESS_CRI_VC_ISSUED.json
mikebeeby: Hmm it's not in the actual events emitted by our integration tests, and it's not in the START schema either (which I copied tbh). 
I guess we could put it in as a non required property of the user object?

Caitlin-cooling: Weird, I wonder if there's something missing from the integration tests because it does get included in the actual events that get emitted when we use a fronted
image

Happy to resolve this comment if we want to raise a ticket to look into this separately as I think it's probably effecting all the schemas (and isn't that big of a deal)

lambdas/issuecredential/src/main/java/uk/gov/di/ipv/cri/address/api/service/VerifiableCredentialService.java
Caitlin-cooling: I have a call about this next week, I _think_ the outcome is going to be that we need to use the correct country codes for crown dependencies but still treat them as UK. 

Let's leave as is for now it's only a small change later if I'm wrong 😅

Caitlin-cooling (approved):
Looks good!

PR #1183, opened 2025-01-27
OJ-3014: OpenTelemtry for TS Lambda

mikebeeby:
### What changed

Add OpenTelemetry to GetAddresses Lambda (Only TS lambda)
Modify Lambda layers in template file so now the TS lambda uses the Node Layer.

### Why did it change

This adds instrumentation to GetAddress TS Lambda so that we can see more details traces in dynatrace.

### Issue tracking
- [OJ-3014](https://govukverify.atlassian.net/browse/OJ-3014)

[OJ-3014]: https://govukverify.atlassian.net/browse/OJ-3014?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/infrastructure/lambda/template.yaml
+++ b/infrastructure/lambda/template.yaml
@@ -103,10 +103,6 @@ Globals:
           - "{{resolve:secretsmanager:${SecretArn}:SecretString:DT_TENANT}}"
           - SecretArn: !FindInMap [EnvironmentConfiguration, !Ref Environment, dynatraceSecretArn]
         DT_OPEN_TELEMETRY_ENABLE_INTEGRATION: true
-    Layers:
-      - !Sub
-        - "{{resolve:secretsmanager:${SecretArn}:SecretString:JAVA_LAYER}}"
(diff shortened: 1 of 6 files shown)


infrastructure/lambda/template.yaml
```diff
@@ -369,6 +369,10 @@ Resources:
           - [!Ref AddressFunctionCanaryErrors]
           - [!Ref AWS::NoValue]
           #- !Ref AddressFunctionCanary5xxErrors
+      Layers:
```
surajkumar, line 372: Can you explain these changes to cloudformation for ts?

surajkumar (changes requested):

infrastructure/lambda/template.yaml
mikebeeby: The Java DT layer was applied to all Lambdas, even though one of our Lambdas is Node. The change is to make the Java lambdas use the Java layer and the Node lambda to use the Node one. 
GetAddresses does not show up in DT traces currently because of this

surajkumar (approved):

PR #1197, opened 2025-02-04
Oj 3037 post postcode lookup endpoint

mikebeeby:
### What changed

**Added POST `/postcode-lookup` endpoint.**

Body example: `{ "postcode": "AA11 1AA" }`

- To change the postcode-lookup endpoint to be a POST request with a body instead of a GET request with the postcode PII in the URL (issues with logging) I have updated the APIGW and existing PostcodeLookup lambda to handle both GET and POST requests.

- Refactored lambda unit tests to include tests for both POST and GET. When GET is removed these tests will need removing and only the POST ones need to remain. The majority of tests only include POST requests.

- Updated the existing integration tests to now use the POST postcode-lookup endpoint. (Note: To ensure GET postcode-lookup endpoint is still tested, added a new feature file which can easily be deleted when GET endpoint is removed)

### Why did it change

We want to create a new private endpoint that our FE can use for searching postcodes, that does not include the users postcode in the URL as this can cause PII issues within logs.

Although this would traditionally be a GET request, the issue with PII in url and how that logs in so many places we decided it needed to be removed from the URL. Of the options, we decided a POST request with the postcode in the body had the best balance of ease of implementation and meeting our requirement.

### Issue tracking
- [OJ-3037](https://govukverify.atlassian.net/browse/OJ-3037)

### Screenshot logs
image
image

[OJ-3037]: https://govukverify.atlassian.net/browse/OJ-3037?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/lambdas/postcode-lookup/src/main/java/uk/gov/di/ipv/cri/address/api/handler/PostcodeLookupHandler.java
+++ b/lambdas/postcode-lookup/src/main/java/uk/gov/di/ipv/cri/address/api/handler/PostcodeLookupHandler.java
@@ -4,6 +4,8 @@ import com.amazonaws.services.lambda.runtime.Context;
 import com.amazonaws.services.lambda.runtime.RequestHandler;
 import com.amazonaws.services.lambda.runtime.events.APIGatewayProxyRequestEvent;
 import com.amazonaws.services.lambda.runtime.events.APIGatewayProxyResponseEvent;
+import com.fasterxml.jackson.core.JsonProcessingException;
+import com.fasterxml.jackson.databind.JsonNode;
 import com.fasterxml.jackson.databind.ObjectMapper;
 import com.nimbusds.oauth2.sdk.ErrorObject;
 import org.apache.logging.log4j.Level;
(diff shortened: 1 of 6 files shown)


surajkumar (changes requested):

jkunle (approved):

lambdas/postcode-lookup/src/main/java/uk/gov/di/ipv/cri/address/api/handler/PostcodeLookupHandler.java
```diff
@@ -111,6 +116,7 @@ public PostcodeLookupHandler(
         this.sessionService = sessionService;
         this.eventProbe = eventProbe;
         this.auditService = auditService;
+        this.objectMapper = new ObjectMapper();
```
jkunle, line 119: I don't think you need this line

surajkumar (approved):

lambdas/postcode-lookup/src/main/java/uk/gov/di/ipv/cri/address/api/handler/PostcodeLookupHandler.java
mikebeeby: I believe I need this for the handlers unit tests, or I'd need to inject a mock Object Mapper. Although, I much prefer using not mocking Object mappers where possible.

jkunle: Ok don't worry, I'll look into it ObjectMapper an indirect dependency of the Handler. I might not need to be mocked

PR #1236, opened 2025-03-06
OJ-2821: Bump cri-lib and re-enable Address5xx Canary

mikebeeby:
### What changed

- Bump cri-lib to 4.1.0 to include the new validateSession logic
- Re-enable `AddressFunctionCanary5xxErrors`
- Set `x-amazon-apigateway-request-validator` to `Validate both` so requests without the `session_id` header return a 400.

### Why did it change

You can get through to the confirm page on a UK journey without a session, this should be handled separately, but we still want to handle this and return a 4XX error instead of a 5XX.

cri-lib `v4.1.0` contains new logic to handle `validateSession` being passed an 'empty' String. Instead of a `DynamoDBException` being returned which is not handled, a `SessionNotFoundException` will be returned.

### Issue tracking
- [OJ-2821](https://govukverify.atlassian.net/browse/OJ-2821)

[OJ-2821]: https://govukverify.atlassian.net/browse/OJ-2821?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -17,7 +17,7 @@ ext {
 		mockito                  : "4.3.1",
 		glassfish_version        : "3.0.3",
 		powertools_version       : "1.18.0",
-		cri_common_lib           : "4.0.0",
+		cri_common_lib           : "4.1.0",
 		pact_provider_version	 : "4.5.11",
 		webcompere_version       : "2.1.6",
 		slf4j_log4j12_version    : "2.0.13", // For contract test debug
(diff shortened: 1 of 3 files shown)


MarieseMikely (approved):
looks good to me!

PR #1265, opened 2025-04-17
Oj 3142 address api test

Christyfemi:
Proposed changes
This API has a rate limit policy documented that means when we send too many requests, it will respond with HTTP status code 429 in the lambda. We wrote a test to cover this scenario.
What changed
The response that we expect from the postcode lookup handler is a 404 HTTP status code indicating that no results could be found with an error message in the response body.
Why did it change
To ensure that endpoints are working correctly and the response HTTP status code is 404 and "Error processing postcode lookup" is generated at the response body.
SCREENSHOT
image

### DIFF
--- a/integration-tests/src/test/java/gov/uk/address/api/stepdefinitions/AddressSteps.java
+++ b/integration-tests/src/test/java/gov/uk/address/api/stepdefinitions/AddressSteps.java
@@ -400,5 +400,6 @@ public class AddressSteps {
     @Then("the response body is Error processing postcode lookup")
     public void theResponseBodyIsErrorProcessingPostcodeLookup() {
         String responseBody = this.testContext.getResponse().body();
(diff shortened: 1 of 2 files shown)


integration-tests/src/test/java/gov/uk/address/api/stepdefinitions/AddressSteps.java
```diff
@@ -75,7 +75,7 @@ public AddressSteps(
         this.addressContext = new AddressContext();
     }
 
-    @When("the user performs a postcode lookup for post code {string}")
+    @And("the user performs a postcode lookup for post code {string}")
```
Caitlin-cooling, line 78: Hmm this change is back again. Does it impact other tests that also use this step?

Christyfemi:
This is ready for PR now

PR #1269, opened 2025-04-24
test: adding a new test steps to the Address API happy path test

Christyfemi:
Proposed changes
The CRIs emit audit events to TXMA (transaction, monitoring and audit) during the journey for monitoring purposes. Events are emitted at crucial stages of the user flow such as when the journey is started, when issuing a verifiable credential and when the journey is ended.

What changed
This test is to add a test coverage that covers the schema validation of the IPV_ADDRESS_CRI_END event which is emitted by the issue credential lambda and preparing a schema for validation. 

Why did it change
We have a test harness which consumes the events from the TXMA queue and saves them to the test-resources-AuditEventsTable-1S2SPCX9ZRYJE database table so that we can test them. 
The main thing is that in the tests we will need to call the test harness events endpoint to get the events for validation.


Christyfemi:
This is ready for review @surajkumar

integration-tests/src/test/resources/features/AddressAPIHappyPath.feature
```diff
@@ -222,3 +222,12 @@ Feature: Address API happy path test
     # TXMA event
     When user sends a GET request to events end point for "IPV_ADDRESS_CRI_VC_ISSUED"
     Then VC_ISSUED TxMA event is validated against schema with isUkAddress "true"
+
+    # schema validation
+    Then user sends a GET request to events end point for "IPV_ADDRESS_CRI_END"
+    Then the IPV_ADDRESS_CRI_END event is emitted and  validated against schema
+
```
surajkumar, line 229: PR looks good! Just a small nitpick but can you remove these additional empty lines? Thanks!

Christyfemi:
the line removed

integration-tests/src/test/resources/schema/IPV_ADDRESS_CRI_END.json
```diff
@@ -0,0 +1,28 @@
...
+      "type": "string"
+    },
+    "component_id": {
+      "type": "string"
+    },
+    "event_name": {
+      "type": "string"
+    },
+    "event_timestamp_ms": {
+      "type": "integer"
```
surajkumar, line 20: timestamps are always `long` datatypes

integration-tests/src/test/resources/schema/IPV_ADDRESS_CRI_END.json
```diff
@@ -0,0 +1,28 @@
...
+    },
+    "timestamp": {
+      "type": "integer"
```
surajkumar, line 23: timestamps are always `long` datatypes

surajkumar (changes requested):
Hi @Christyfemi, the schema you have created is incorrect. Please could you double check it? You can use the START event schema as a reference point. 

Thanks

integration-tests/src/test/resources/schema/IPV_ADDRESS_CRI_END.json
Christyfemi: @surajkumar timestamps should be integer based on https://event-catalogue.internal.account.gov.uk/events/IPV_ADDRESS_CRI_END/

integration-tests/src/test/resources/schema/IPV_ADDRESS_CRI_END.json
```diff
@@ -0,0 +1,28 @@
+{
+  "type": "object",
+  "properties": {
```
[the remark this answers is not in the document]
Christyfemi: I will connect with you tomorrow to clarify the issue. There is no persistent_session_id among the field specified here-https://event-catalogue.internal.account.gov.uk/events/IPV_ADDRESS_CRI_END/

integration-tests/src/test/resources/schema/IPV_ADDRESS_CRI_END.json
surajkumar: That's interesting... 

An int is a 32-bit value with a max of value 2147483647, once you go past that value, we get into the realm of integer overflow i.e. the time becomes `-2147483647`. 

Given this, an integer can't even hold the current time in millis. When the data type is a long we can go all the way up to the year 292 billion

surajkumar: Basically, what I'm saying is that the doc you've linked is incorrect as we physically can't use an integer

Christyfemi:
This is now passed. @surajkumar

Christyfemi:
All done and check at your convenience

surajkumar (approved):

PR #1292, opened 2025-06-06
OJ-3140 - Set up `NoMatchingEncryptionKeyAlarm`

barnabycollins:
### What changed

- Defined new `NoMatchingEncryptionKeyAlarm`

### Why did it change

We want to know when the CRI fails to decrypt a JWT with any of the available key aliases

### Issue tracking

OJ-3140

- [x] No environment variables or secrets were added or changed

### Other considerations

- [x] Update [README](./blob/main/README.md) with any new instructions or tasks

### DIFF
--- a/infrastructure/lambda/template.yaml
+++ b/infrastructure/lambda/template.yaml
@@ -53,10 +53,12 @@ Conditions:
   UseSecretPrefix: !Not [!Equals [!Ref SecretPrefix, "none"]]
   UsePermissionsBoundary: !Not [!Equals [!Ref PermissionsBoundary, "none"]]
   UseCodeSigningConfig: !Not [!Equals [!Ref CodeSigningConfigArn, "none"]]
-  UseCanaryDeploymentAlarms: !Not [!Equals [!Ref LambdaDeploymentPreference, AllAtOnce]]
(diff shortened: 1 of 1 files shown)


infrastructure/lambda/template.yaml
```diff
@@ -854,10 +854,10 @@ Resources:
       ComparisonOperator: GreaterThanThreshold
       TreatMissingData: notBreaching
 
-  NoMatchingEncryptionKeyAlarm:
+  SessionLambdaNoMatchingEncryptionKeyAlarm:
     Type: AWS::CloudWatch::Alarm
     Properties:
-      AlarmName: !Sub "${AWS::StackName}-${Environment}-NoMatchingEncryptionKeyAlarm"
+      AlarmName: !Sub "${AWS::StackName}-${Environment}-SessionLambdaNoMatchingEncryptionKeyAlarm"
```
surajkumar, line 860: nit' but we haven't been setting names for alarms explicitly and I don't think we need to do so now?

surajkumar (changes requested):

infrastructure/lambda/template.yaml
surajkumar: Ignore - we have for the jwt stuff

surajkumar (approved):

PR #1295, opened 2025-06-11
OJ-3214: Change SupportManualURL to mapping

Caitlin-cooling:
### What changed

- Change SupportManualURL to mapping
- Add SupportManualURL to alarms where it wasn't present

### Why did it change

Because params can't be easily updated

### Screenshots

Before
image

After

### Issue tracking

- [OJ-3214](https://govukverify.atlassian.net/browse/OJ-3214)

[OJ-3214]: https://govukverify.atlassian.net/browse/OJ-3214?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/infrastructure/lambda/template.yaml
+++ b/infrastructure/lambda/template.yaml
@@ -44,10 +44,6 @@ Parameters:
     Description: "The stack containing the TXMA infrastructure"
     Type: String
     Default: txma-infrastructure
-  SupportManualURL:
-    Type: String
-    Default: "https://team-manual.account.gov.uk/teams/CRI-Orange-team/supporting-cri-orange/address-credential-issuer-runbook"
-    Description: The support manual URL to be provided in alarm messages.
 
 Conditions:
   UseSecretPrefix: !Not [!Equals [!Ref SecretPrefix, "none"]]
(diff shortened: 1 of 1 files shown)


infrastructure/lambda/template.yaml
```diff
@@ -191,6 +187,10 @@ Globals:
         }
 
 Mappings:
+  StaticVariables:
+    Urls:
+      SupportManualURL: "https://team-manual.account.gov.uk/teams/CRI-Orange-team/supporting-cri-orange/address-credential-issuer-runbook"
```
barnabycollins, line 192: Should this include the anchor link to the relevant section? Indifferent either way, just thought the question was worth asking as it could move down the page at some point in the future

Caitlin-cooling: Yeah this support page unfortunately doesn't have nice anchors splitting the alarms into FE and BE so the anchor would just go to the top of the page

barnabycollins: Yeah, fair enough

PR #1299, opened 2025-06-16
OJ-3165: feat - add retry the to addressItem logic to prevent null po…

MarieseMikely:
…inter exceptions

### Issue tracking

- [OJ-XXXX](https://govukverify.atlassian.net/browse/OJ-XXX)

### Other considerations


jkunle:
Looking the error message looks like it gets as far as https://github.com/govuk-one-login/ipv-cri-address-api/blob/main/lambdas/issuecredential/src/main/java/uk/gov/di/ipv/cri/address/api/handler/IssueCredentialHandler.java#L143 which means potentially 1 of 2 things is causing it to throw a `NullPointerException` i.e

either `sessionItem.getSubject()` or it could be `addressItem.getAddresses()` so this retry, I think may not make much of a difference, the reason is the code seems to get beyond `line 139`

I am more suspicious of `sessionItem.getSubject()` b'cos we retrieve var sessionItem = this.sessionService.getSessionByAccessToken(accessToken);` meaning you might have a situation where the is `partial object` `sessionItem` object, one for which when you call `session.getSubject()` it appears there is `no subject` and hence the null pointer exception
 
So maybe we should use `access-token-index-with-event-data` which has a higher guarantee of bringing in the `subject 

see `https://github.com/govuk-one-login/ipv-cri-common-lambdas/blob/main/infrastructure/lambda/template.yaml#L1317`

jkunle (dismissed):

lambdas/issuecredential/src/main/java/uk/gov/di/ipv/cri/address/api/handler/IssueCredentialHandler.java
```diff
@@ -246,4 +265,25 @@ private AccessToken validateInputHeaderBearerToken(Map<String, String> headers)
...
+
+    private AddressItem getAddressItemsWithRetries(SessionItem sessionItem) {
+        RetryConfig retryConfig = getRetryConfig(500, 3, true);
+        try {
+            return RetryManager.execute(
+                    retryConfig, () -> addressService.getAddressItem(sessionItem));
+        } catch (AddressNotFoundException e) {
+            eventProbe.log(
+                    Level.WARN,
+                    "Initial get address item returned null. Going through retry logic..");
```
mikebeeby, line 285: Is this log correct, is this not after retries?

lambdas/issuecredential/src/main/java/uk/gov/di/ipv/cri/address/api/handler/IssueCredentialHandler.java
```diff
@@ -246,4 +265,25 @@ private AccessToken validateInputHeaderBearerToken(Map<String, String> headers)
...
+
+    private RetryConfig getRetryConfig(int delayMs, int maxAttempts, boolean exponential) {
+        return new RetryConfig.Builder()
+                .delayBetweenAttempts(delayMs)
+                .maxAttempts(maxAttempts)
+                .exponentiallyRetry(exponential)
+                .build();
+    }
+
+    private AddressItem getAddressItemsWithRetries(SessionItem sessionItem) {
```
mikebeeby, line 277: Should this and the retry logic not be in the AddressService?

lambdas/issuecredential/src/main/java/uk/gov/di/ipv/cri/address/api/handler/IssueCredentialHandler.java
```diff
@@ -136,7 +142,8 @@ public APIGatewayProxyResponseEvent handleRequest(
             var sessionItem = this.sessionService.getSessionByAccessToken(accessToken);
 
             eventProbe.log(Level.INFO, "found session");
-            var addressItem = addressService.getAddressItem(sessionItem.getSessionId());
+
+            AddressItem addressItem = getAddressItemsWithRetries(sessionItem);
```
mikebeeby, line 146: Not sure this really needs to be called `..WithRetries`? 
Also should it be getAddressItem still?

lib/src/main/java/uk/gov/di/ipv/cri/address/library/service/AddressService.java
```diff
@@ -90,7 +92,19 @@ public AddressItem saveAddresses(
     }
 
     public AddressItem getAddressItem(UUID sessionId) {
-        return dataStore.getItem(String.valueOf(sessionId));
+        return dataStore.getItem(sessionId.toString());
+    }
+
+    public AddressItem getAddressItem(SessionItem sessionItem) {
```
mikebeeby, line 98: This only needs the sessionId, not the full item. Not sure overloading this method like this is the best approach. Why are we not just changing the existing method?

mikebeeby (approved):

PR #1305, opened 2025-06-20
Add task name check

Caitlin-cooling:
### What changed

Adds a check for `jacocoTestReport` tasks

### Why did it change

Java unit test coverage was not being genarated


surajkumar:
Good catch @Caitlin-cooling !!

PR #1325, opened 2025-08-04
OJ-3313: Save Trimmed Postcode

jkunle:
## Proposed changes

DVLA has reported validation issues due to inconsistent formatting of postcodes, submitted via our address entry field this include  postcodes containing extra or irregular white spaces this is as expected see https://design-system.service.gov.uk/patterns/addresses/

However, when user has committed to address changes; the post code should be stored in a way as to prevent  reported validation issues

### What changed

Store postcode with any validation issues

**Postcode with trailing space and extra space in the middle**

image

image

image

image

**Click on Change link**

image

**Confirm your details appear like this**
image

**Click on all change link reveals - post code is still with spaces**
image

Postcode in dynamodb is with spaces

image

**NB:**
Added a scenario in `AddressAPIHappyPath.feature` called `uk_address_save_address_remove_spaces_from_postcode`
This implements a manual UK address flow i.e. UK postcode with spaces has been manually entered. This test saving addresses with different passcode values with spaces and compares the postcode retrieved in the generated VC output with the expected (non spaced) postcode from the specified example see below

image

**DVLA** is for England so -> Guernsey, Jersey and Isle of Man are out of scope
### Issue tracking

- [OJ-3313](https://govukverify.atlassian.net/browse/OJ-3313)

[OJ-3313]: https://govukverify.atlassian.net/browse/OJ-3313?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/lib/src/main/java/uk/gov/di/ipv/cri/address/library/service/AddressService.java
+++ b/lib/src/main/java/uk/gov/di/ipv/cri/address/library/service/AddressService.java
@@ -7,7 +7,6 @@ import com.fasterxml.jackson.databind.ObjectReader;
 import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
 import software.amazon.awssdk.enhanced.dynamodb.DynamoDbEnhancedClient;
-import software.amazon.awssdk.utils.StringUtils;
 import uk.gov.di.ipv.cri.address.library.exception.AddressNotFoundException;
 import uk.gov.di.ipv.cri.address.library.exception.AddressProcessingException;
 import uk.gov.di.ipv.cri.address.library.persistence.item.AddressItem;
@@ -25,6 +24,9 @@ import java.util.List;
 import java.util.Objects;
 import java.util.Optional;
 import java.util.UUID;
+
+import static uk.gov.di.ipv.cri.address.library.util.CountryCode.isCountryCodeAbsentForAny;
+import static uk.gov.di.ipv.cri.address.library.util.CountryCode.isGreatBritain;
 
 public class AddressService {
     private static final Logger LOGGER = LogManager.getLogger();
@@ -76,7 +78,7 @@ public class AddressService {
             throw new AddressProcessingException("could not parse addresses..." + e.getMessage());
         }
 
-        if (!isCountryCodePresentForAll(addresses)) {
+        if (isCountryCodeAbsentForAny(addresses)) {
             throw new AddressProcessingException(ERROR_COUNTRY_CODE_NOT_PRESENT);
         }
         return addresses;
@@ -90,7 +92,7 @@ public class AddressService {
         addressItem.setAddresses(
                 Optional.ofNullable(addresses).orElse(Collections.emptyList()).stream()
                         .filter(Objects::nonNull)
-                        .map(this::normalizePostCodeAddresses)
+                        .map(this::normalizePostcodeUkAddresses)
                         .toList());
         dataStore.create(addressItem);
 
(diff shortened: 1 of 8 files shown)


lib/src/main/java/uk/gov/di/ipv/cri/address/library/service/AddressService.java
```diff
@@ -96,6 +102,22 @@ public AddressItem saveAddresses(
         return addressItem;
     }
 
+    private CanonicalAddress normalizePostCodeAddresses(CanonicalAddress address) {
+        String rawPostcode = address.getPostalCode();
+        if (rawPostcode != null && "GB".equalsIgnoreCase(address.getAddressCountry())) {
```
Caitlin-cooling, line 107: This is interesting, we're going to need to update the condition when we make this change https://govukverify.atlassian.net/browse/OJ-3105. I wonder if we want something like a `isUKAddress` helper that can be reused and easily updated in one place so that we don't miss things like this?

lib/src/test/java/uk/gov/di/ipv/cri/address/library/service/AddressServiceTest.java
```diff
@@ -319,6 +323,72 @@ void shouldPersistAnEmptyListOfAddressesWhenNoListOfCanonicalAddressesIsSupplied
...
+        @CsvSource({
+            "'M1 1AA', 'M11AA'", // correct
+            "'B33 8TH', 'B338TH'", // correct
+            "'W1A0AX', 'W1A0AX'", // correct
+            "'SE1 9GP', 'SE19GP'", // already correct
+            "' E1 6AN', 'E16AN'", // should remove leading space
+            "'N1 9GU ', 'N19GU'", // should remove trailing space
+            "' W2 1HB ', 'W21HB'", // should remove leading and trailing spaces
+            "'SW1A  2AA', 'SW1A2AA'", // should normalize double space to single
+            "'   M25  4RT', 'M254RT'" // should remove leading spaces and normalize double space
```
Caitlin-cooling, line 337: Great test coverage and acts as really nice documentation thanks Ola!

lib/src/main/java/uk/gov/di/ipv/cri/address/library/service/AddressService.java
```diff
@@ -96,6 +102,22 @@ public AddressItem saveAddresses(
...
+
+            String formattedPostCode = rawPostcode.replace(" ", "").toUpperCase();
+            int postCodeLength = formattedPostCode.length();
+            int minUKPostcodeLength = 5;
+            int maxUKPostcodeLength = 7;
+
+            if (postCodeLength >= minUKPostcodeLength && postCodeLength <= maxUKPostcodeLength) {
+                address.setPostalCode(formattedPostCode);
```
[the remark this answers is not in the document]
Caitlin-cooling: Ah ok! If it's left over and we already validate more thoroughly what do you think about just removing it?

lib/src/main/java/uk/gov/di/ipv/cri/address/library/service/AddressService.java
jkunle: I have included the regression element on **AC 4** I added a CountryCode util class this also moving the logic to one location. I believe with this changes this should be missed by anyone who picks this up

Caitlin-cooling (approved):

PR #1327, opened 2025-08-12
OJ-3274: Add OS API URL SSM params per client

mikebeeby:
### What changed

Adds Ordnance Survery API URL SSM params for different clients, example SSM param name: `"/${AWS::StackName}/OrdnanceSurveyAPIUrl/ipv-core"`

Fixed gradle version issue in GHA's, original KBV PR here: https://github.com/govuk-one-login/ipv-cri-kbv-api/pull/534

### Why did it change

Create the SSM params to allow (in future PRs) Core and other consumers to dynamically configure whether the stub or real API should be used, depending on the Client ID provided 

### Issue tracking

- [OJ-3274](https://govukverify.atlassian.net/browse/OJ-3274)

[OJ-3274]: https://govukverify.atlassian.net/browse/OJ-3274?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/infrastructure/lambda/template.yaml
+++ b/infrastructure/lambda/template.yaml
@@ -56,6 +56,7 @@ Conditions:
   IsLocalDevEnvironment: !Equals [!Ref Environment, localdev]
   IsBuildEnvironment: !Equals [!Ref Environment, build]
   IsProdEnvironment: !Equals [!Ref Environment, production]
+  IsNotProdEnvironment: !Not [!Equals [!Ref Environment, production]]
   IsNotDevEnvironment: !Not [!Condition IsDevEnvironment]
   IsDevLikeOrBuild: !Or
     - !Condition IsDevEnvironment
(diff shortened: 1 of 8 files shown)


infrastructure/lambda/template.yaml
```diff
@@ -753,6 +763,110 @@ Resources:
...
+
+  IpvCoreStubAwsBuildOrdnanceSurveyAPIURL:
+    Condition: IsNotProdEnvironment
+    Type: AWS::SSM::Parameter
+    Properties:
+      Name: !Sub "/${AWS::StackName}/OrdnanceSurveyAPIUrl/ipv-core-stub-aws-build"
+      Type: String
+      Value: !FindInMap [ StubOrdnanceSurveyAPIURLs, Environment, !Ref Environment ]
+
+  IpvCoreStubOrdnanceSurveyAPIURL:
```
jkunle, line 806: do we actually need this, the old PASS client Id?

jkunle: Ummh, are you sure? I suppose there is no damage to leave it in

PR #1334, opened 2025-08-19
OJ-3115: Reject dates on address endpoint with year 0000

mikebeeby:
### What changed

Adds regex to reject `0000` as the date year for `validFrom` and `validTo` in the API Address schema. 
Added an integration test to test this.

### Why did it change

If a request is sent to our `/address` with a `validFrom` or `validUntil` year of `0000` the deserialisation will error as LocalDate does not see `0000` as a valid date year. 

We have implemented a fix FE https://github.com/govuk-one-login/ipv-cri-address-front/pull/1209 but we also want to harden our endpoint to prevent `0000` causing `5XX` errors.

### Issue tracking

- [OJ-3115](https://govukverify.atlassian.net/browse/OJ-3115)

[OJ-3115]: https://govukverify.atlassian.net/browse/OJ-3115?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/integration-tests/src/test/resources/features/AddressUnhappyPath.feature
+++ b/integration-tests/src/test/resources/features/AddressUnhappyPath.feature
@@ -41,3 +41,27 @@ Feature: Address API unhappy path test
     Given a request is made to the postcode-lookup endpoint without a postcode in the body and with session id in the header
     Then the endpoint should return a 400 HTTP status code
     And the response body contains no postcode error
+
+  @invalid_date
+  Scenario: Date in the request body is the invalid value 0000-01-01
+    Given a request is made to the address endpoint with a validFrom date of "0000-01-01"
+    Then the endpoint should return a 400 HTTP status code
+    And the response body contains bad request error
+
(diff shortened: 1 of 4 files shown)


integration-tests/src/test/resources/features/AddressUnhappyPath.feature
```diff
@@ -41,3 +41,9 @@ Feature: Address API unhappy path test
     Given a request is made to the postcode-lookup endpoint without a postcode in the body and with session id in the header
     Then the endpoint should return a 400 HTTP status code
     And the response body contains no postcode error
+
+  @invalid_date
+  Scenario: Date year in the request body is the invalid value 0000
```
Caitlin-cooling, line 46: This is just one part of the regex, right? Shall we also add a test for the other parts too?

PR #1459, opened 2026-07-28
feat: add optional X-Forwarded-For header to session endpoint

ulrictene-ops:
### What changed
Added the optional X-Forwarded-For header parameter to the /session endpoint in the Address CRI private OpenAPI specification.

### Why did it change
To align the Address CRI session endpoint header configuration with the agreed standard used by Check HMRC. This ensures a consistent API contract across CRIs and removes the need for consumers to implement CRI-specific handling for the X-Forwarded-For header.

### Issue tracking

- [OJ-3801](https://govukverify.atlassian.net/browse/OJ-3801)

[OJ-3801]: https://govukverify.atlassian.net/browse/OJ-3801?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/infrastructure/lambda/private-api.yaml
+++ b/infrastructure/lambda/private-api.yaml
@@ -107,14 +107,15 @@ paths:
 
   /session:
     post:
+      summary: "IP address of the client."
       parameters:
+        - in: header
+          name: "X-Forwarded-For"
+          schema:
+            type: string
+            format: string
+            required: false
         - $ref: "#/components/parameters/AuditHeader"
-      requestBody:
-        content:
-          application/json:
-            schema:
-              $ref: "#/components/schemas/Authorization"
-        required: true
       responses:
         "400":
           description: "400 response"
--- a/infrastructure/lambda/private-api.yaml
+++ b/infrastructure/lambda/private-api.yaml
@@ -107,15 +107,21 @@ paths:
 
   /session:
     post:
-      summary: "IP address of the client."
       parameters:
         - in: header
           name: "X-Forwarded-For"
+          description: "IP address of the client"
           schema:
             type: string
             format: string
             required: false
         - $ref: "#/components/parameters/AuditHeader"
+      requestBody:
+        content:
+          application/json:
+            schema:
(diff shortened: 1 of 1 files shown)


infrastructure/lambda/private-api.yaml
```diff
@@ -107,14 +107,15 @@ paths:
...
+            type: string
+            format: string
+            required: false
         - $ref: "#/components/parameters/AuditHeader"
-      requestBody:
-        content:
-          application/json:
-            schema:
-              $ref: "#/components/schemas/Authorization"
-        required: true
```
barnabycollins, line 117: I don't think we want to get rid of this - the authorization endpoint needs to retain the request body spec as well as gaining the new optional entry for the header. For reference, the `#/components/schemas/Authorization` entry can be found around line 273 in the file 👍

infrastructure/lambda/private-api.yaml
```diff
@@ -107,14 +107,15 @@ paths:
 
   /session:
     post:
+      summary: "IP address of the client."
```
barnabycollins: This summary is for the `POST /session` endpoint, not just for the header. Looking at the [reference](https://swagger.io/specification/#common-fixed-fields), it's possible to add a `description` field under the header itself, so that might be preferable to a `summary` on the whole endpoint. I'll add a code suggestion to show what I think we can do :)

infrastructure/lambda/private-api.yaml
```diff
@@ -107,14 +107,15 @@ paths:
...
       parameters:
+        - in: header
+          name: "X-Forwarded-For"
```
barnabycollins, line 112:
```suggestion
          name: "X-Forwarded-For"
          description: "IP address of the client"
```

barnabycollins (commented):
Thanks for this Ulric. Just a few comments and then I think we'll be good :)

barnabycollins (approved):

